PR #2, opened 2018-12-13
Data structure for the app

webbyfox:


djangocms_navigation/models.py
```diff
@@ -0,0 +1,58 @@
...
+    pass
+
+
+class MenuContent(models.Model):
+    title = models.CharField(verbose_name=_('title'), max_length=100)
+    site = models.ForeignKey(
+        Site,
+        on_delete=models.PROTECT,
+    )
+    menu = models.OneToOneField(
```
monikasulik: ForeignKey not OneTo One! :)

djangocms_navigation/admin.py
```diff
@@ -0,0 +1,29 @@
...
+from treebeard.forms import movenodeform_factory
+
+from .models import Menu, MenuContent, MenuItem
+
+
+class MenuContentAdmin(admin.ModelAdmin):
+    exclude = ['modified_by', 'menu', ]
+
+    def save_model(self, request, obj, form, change):
+        if not obj.pk:
```
monikasulik: After changing the one2one to an FK, this will need to be `if not obj.menu`.

djangocms_navigation/admin.py
```diff
@@ -0,0 +1,29 @@
...
+        super().save_model(request, obj, form, change)
+
+
+class MenuItemAdmin(TreeAdmin):
+    form = movenodeform_factory(MenuItem)
+
+    exclude = ['modified_by', ]
+
+    def save_model(self, request, obj, form, change):
+        obj.modified_by = request.user
```
monikasulik: As discussed, I think either maybe we want to move modified_by to versioning or remove it (if it's not possible to ever have somebody else modify a version other than the person who created it) or do `obj.menu_content.modified_by = request.user`

djangocms_navigation/models.py
```diff
@@ -0,0 +1,58 @@
...
+        Menu,
+        on_delete=models.CASCADE,
+        primary_key=True,
```
monikasulik: so if it's an FK it can't be a PK anymore.
Btw, not something you can use here (because it shouldn't really be a one2one), but for future reference, when you have this kind of db structure, this is a pretty cool django feature: https://docs.djangoproject.com/en/2.1/topics/db/models/#multi-table-inheritance

djangocms_navigation/models.py
```diff
@@ -0,0 +1,58 @@
...
+class MenuItem(MP_Node):
+    title = models.CharField(verbose_name=_('title'), max_length=100)
+    content_type = models.ForeignKey(
+        ContentType,
+        on_delete=models.PROTECT,
+    )
+    object_id = models.PositiveIntegerField()
+    content = GenericForeignKey('content_type', 'object_id')
+    menu_content = models.ForeignKey(
+        MenuContent, on_delete=models.CASCADE
```
monikasulik: We discussed possibly doing this with PROTECT instead.

djangocms_navigation/models.py
```diff
@@ -0,0 +1,58 @@
...
+
+class MenuContent(models.Model):
+    title = models.CharField(verbose_name=_('title'), max_length=100)
+    site = models.ForeignKey(
+        Site,
+        on_delete=models.PROTECT,
+    )
+    menu = models.OneToOneField(
+        Menu,
+        on_delete=models.CASCADE,
```
monikasulik: We discussed possibly doing this with PROTECT instead.

djangocms_navigation/models.py
```diff
@@ -0,0 +1,58 @@
...
+        primary_key=True,
+    )
+    modified_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.PROTECT,
+        verbose_name=_('modified_by'),
+        null=True,
```
monikasulik: Should this ever be null? If we're doing PROTECT on delete, then I'm not sure there's a reason to have null=True actually.

djangocms_navigation/admin.py
webbyfox: This still won't work for obj.menu. django will throw `RelatedObjectDoesNotExist` on check. I will keep obj.pk here.

djangocms_navigation/admin.py
```diff
@@ -0,0 +1,21 @@
...
+from treebeard.forms import movenodeform_factory
+
+from .models import Menu, MenuContent, MenuItem
+
+
+class MenuContentAdmin(admin.ModelAdmin):
+    exclude = ['menu', ]
+
+    def save_model(self, request, obj, form, change):
+        if not obj.id:
```
monikasulik: Maybe add a comment about why we're doing this here. It's the sort of thing that could confuse somebody

chaosk: `if not change`?

djangocms_navigation/models.py
```diff
@@ -0,0 +1,45 @@
...
+class Menu(models.Model):
+    """
+    MenuContent Grouper
+    """
+    pass
+
+
+class MenuContent(models.Model):
+    title = models.CharField(verbose_name=_('title'), max_length=100)
+    site = models.ForeignKey(
```
chaosk: I still don't know why this relation is here and not in Menu. Would it mean you can just switch an existing menu to a different site in a different version? Does it mean you can also have two published menu contents that point to a single site?

djangocms_navigation/models.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class MenuItem(MP_Node):
+    title = models.CharField(verbose_name=_('title'), max_length=100)
+    content_type = models.ForeignKey(
+        ContentType,
+        on_delete=models.PROTECT,
+    )
+    object_id = models.PositiveIntegerField()
+    content = GenericForeignKey('content_type', 'object_id')
```
chaosk: Do we need a separate field for external URL, or is it covered by URL manager?

chaosk: What about the 'open in new window' setting?

djangocms_navigation/models.py
webbyfox: My thinking was, a single site can have multiple menus. I raised question with Noel to confirm

djangocms_navigation/models.py
chaosk: But that won't work if you're pointing to a Page here, for example.

djangocms_navigation/models.py
chaosk: I think site and menu identifier should be in Menu then, and they should be uneditable/unique_together.

monikasulik: That's assuming we shouldn't be able to change the site after adding a menu (or if we do change it then the change should not be versioned). It would make sense to me that menus shouldn't suddenly change sites, but I'm not sure what the requirement is tbh.
And my understanding is the same as Riz's - that we can have multiple menus on one site. So for example there could be a top nav menu at the top which maybe all pages would have, but then a sidebar menu which changes depending on the section of the site we're in.

djangocms_navigation/models.py
webbyfox: I am thinking it now and it would be easier to add another option rather rely on model. Will add

djangocms_navigation/models.py
webbyfox: Now I am thinking it and I agree site should be in Menu however it shouldn't be unique_together to allow multiple. For example, if user created menu under one site and plugin render will use that particular site and all the node added to generate URL but if then site change then all node URL won't work as they belong to another site, right?

chaosk: That's why you get unique_together on (site, identifier). This way you can't have two menus with the same identifier on the same site.

djangocms_navigation/models.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    MenuContent Grouper
+    """
+    site = models.ForeignKey(
+        Site,
+        on_delete=models.PROTECT,
+    )
+
+    class Meta:
+        unique_together = ('id', 'site')
```
chaosk:
```suggestion
        unique_together = (('id', 'site'), )
```

PR #6, opened 2019-01-02
Added prototype

webbyfox:


djangocms_navigation/admin.py
```diff
@@ -19,9 +22,84 @@ def save_model(self, request, obj, form, change):
             )
         super().save_model(request, obj, form, change)
 
+    def get_list_display(self, request):
```
chaosk: If you're not adding any logic to modify the contents of list_display, consider just setting list_display attribute

djangocms_navigation/admin.py
```diff
@@ -19,9 +22,84 @@ def save_model(self, request, obj, form, change):
...
+        list_display = ["title", "get_menuitem_link"]
+        return list_display
+
+    def get_menuitem_link(self, obj):
+        object_preview_url = reverse(
+            "admin:{app}_menuitem_list".format(
```
chaosk:
```suggestion
             "admin:{app}_{model}_list".format(
```

djangocms_navigation/admin.py
```diff
@@ -19,9 +22,84 @@ def save_model(self, request, obj, form, change):
...
+                app=obj._meta.app_label, model=obj._meta.model_name
+            ),
+            args=[obj.pk],
+        )
+
+        return format_html(
+            '<a href="{}" class="js-moderation-close-sideframe" target="_top">'
+            '<span class="cms-icon cms-icon-eye"></span> Items'
```
chaosk: Make "Items" translatable

djangocms_navigation/admin.py
```diff
@@ -19,9 +22,84 @@ def save_model(self, request, obj, form, change):
...
+
+    def get_queryset(self, request):
+        if hasattr(self, "menu_content_id"):
+            return MenuItem.objects.filter(menu_content=self.menu_content_id)
+        return self.model().get_tree()
+
+    def add_view(self, request, menu_content_id=None, form_url="", extra_context=None):
+        extra_context = extra_context or {}
+        if menu_content_id:
+            self.menu_content_id = menu_content_id
```
chaosk: Can you set this on request instead?

djangocms_navigation/admin.py
webbyfox: the model here is 'menucontent` hence I hardcorded to `menuitem`

PR #8, opened 2019-01-03
Admin Implementation for navigation tree

webbyfox:


djangocms_navigation/forms.py
```diff
@@ -1,9 +1,71 @@
...
+            choices.append(
+                (
+                    ContentType.objects.get_for_model(model).id,
+                    model._meta.verbose_name.capitalize(),
+                )
+            )
+
+            content_choices.extend([(obj.id, obj) for obj in model.objects.all()])
+
+        self.fields["content_type"].choices = choices
```
monikasulik: This seems quite complex for what it does really. Wondering why we're not just using `ModelChoiceField` and then doing something like `self.fields["content_type"].queryset = supported_models()` without all this looping and appending?

djangocms_navigation/forms.py
```diff
@@ -1,9 +1,71 @@
...
+        content_choices = []
+        for model in supported_models():
+            choices.append(
+                (
+                    ContentType.objects.get_for_model(model).id,
+                    model._meta.verbose_name.capitalize(),
+                )
+            )
+
+            content_choices.extend([(obj.id, obj) for obj in model.objects.all()])
```
monikasulik: Won't this mean that the choice list could potentially be veeeeeeery long? And won't that potentially lead to the browser crashing (I've seen it happen)? Perhaps discuss with @chaosk ? Might make sense to have this field as an autocomplete or something like that. There are django libs that add autocomplete to the admin, so hopefully won't be too complicated, but maybe best done in a separate ticket.

djangocms_navigation/forms.py
```diff
@@ -1,9 +1,71 @@
...
+
+        self.fields["content_type"].choices = choices
+        self.fields["object_id"].choices = content_choices
+
+        if self.instance:
+
+            if self.instance.content_type_id:
```
monikasulik: How come this doesn't happen automatically? I would have thought a model form would always do this?

djangocms_navigation/forms.py
```diff
@@ -1,9 +1,71 @@
...
+                self.fields["content_type"].initial = self.instance.content_type
+                self.fields["object_id"].initial = self.instance.object_id
+
+    def clean(self):
+        data = super().clean()
+        data["content_type"] = ContentType.objects.get(id=data["content_type"])
+        return data
```
monikasulik: This looks quite weird. Why doesn't this happen automatically?

djangocms_navigation/forms.py
```diff
@@ -1,9 +1,71 @@
...
+
+    def save(self, **kwargs):
+        self.instance.object_id = self.cleaned_data["object_id"]
+        self.instance.content_type = self.cleaned_data["content_type"]
```
monikasulik: And again?? If there's something in this class which is causing the django model form to not pick up these fields I think it's worth putting in some comments in the code as to why.

djangocms_navigation/models.py
```diff
@@ -28,8 +28,7 @@ class Meta:
 
 class MenuContent(models.Model):
     menu = models.ForeignKey(Menu, on_delete=models.PROTECT)
-    root = models.ForeignKey(
-        'djangocms_navigation.MenuItem', on_delete=models.PROTECT)
+    root = models.ForeignKey("djangocms_navigation.MenuItem", on_delete=models.PROTECT)
```
monikasulik: Technically speaking you've changed it from pep8 compliant to non-pep8 compliant - just sayin' ;) I know max line length isn't a big deal and there are some cases where to make the line short is to make things very difficult to read, but sometimes when doing merges in split screen etc. it's actually useful to have the lines under 80 chars.

djangocms_navigation/test_utils/factories.py
```diff
@@ -13,10 +13,11 @@
 
 class UserFactory(factory.django.DjangoModelFactory):
     username = FuzzyText(length=12)
-    first_name = factory.Faker('first_name')
-    last_name = factory.Faker('last_name')
+    first_name = factory.Faker("first_name")
```
monikasulik: hmmm, why all these changes from `'` to `"`? Both are correct, but generally most of the time when the strings are short like here people tend to use `'` and when they're sentences `"` is used more commonly.

djangocms_navigation/utils.py
```diff
@@ -0,0 +1,26 @@
+import functools
+
```
monikasulik: This is definitely not pep8 standard :P Both of the functools lines are standard libs so should be grouped together really

djangocms_navigation/utils.py
```diff
@@ -0,0 +1,26 @@
...
+from django.apps import apps
+
+
+@lru_cache(maxsize=1)
+def supported_models():
+    app_config = apps.get_app_config("djangocms_navigation")
+    try:
+        extension = app_config.cms_extension
+        return extension.navigation_apps_models
+    except AttributeError:
```
monikasulik: Why is this necessary?

djangocms_navigation/utils.py
```diff
@@ -0,0 +1,26 @@
+import functools
+
+from functools import lru_cache
+
+from django.apps import apps
+
+
+@lru_cache(maxsize=1)
+def supported_models():
```
monikasulik: Wouldn't it make more sense if this returned a queryset of content type objects? I mean otherwise this is just an alias for `extension.navigation_apps_models`, no?

djangocms_navigation/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    app_config = apps.get_app_config("djangocms_navigation")
+    try:
+        extension = app_config.cms_extension
+        return extension.navigation_apps_models
+    except AttributeError:
+        return app_config.navigation_apps_models
+
+
+def is_model_supported(model):
```
monikasulik: This doesn't appear to be used anywhere so why is it here?

djangocms_navigation/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    """Return bool value if model is in keys"""
+    return model in supported_models().keys()
+
+
+def get_supported_model_queryset(model):
```
monikasulik: Ditto, not used anywhere and not sure what it does really. I don't think what it returns is actually a queryset?

djangocms_navigation/models.py
webbyfox: I use black to format code and max line length is taken from setup.cfg which is currently 120 for all our projects.

djangocms_navigation/admin.py
```diff
@@ -1,26 +1,52 @@
...
 
 class MenuContentAdmin(admin.ModelAdmin):
-    exclude = ["menu"]
+    form = MenuContentForm
     list_display = ["title", "get_menuitem_link"]
+    list_display_links = None
 
     def save_model(self, request, obj, form, change):
         if not change:
+            title = form.data.get("title")
```
monikasulik: I think I missed this before... This is actually potentially a security risk - `form.data` is the data as it came in the POST. What you want to be getting is `form.cleaned_data`.

monikasulik:
@webbyfox Are tests passing for you? Failing for me.

webbyfox:
@monikasulik , Yes I am running `python setup.py test` and ran OK

djangocms_navigation/utils.py
```diff
@@ -0,0 +1,13 @@
+from functools import lru_cache
+
+from django.apps import apps
+
+
+@lru_cache(maxsize=1)
+def supported_models():
+    try:
+        app_config = apps.get_app_config("djangocms_navigation")
```
monikasulik: It is generally bad practice to put multiple lines into a try clause unless you have a good reason for it. What can happen is that some line we weren't expecting to throw an exception does and then it gets caught and then someone spends ages debugging (have had this happen on my teams more than once). In this particular case `LookupError` is exotic enough that it's unlikely, but still good to get into good habits. AFAIK, what you probably want here is:
```
try:
    app_config = apps.get_app_config("djangocms_navigation")
except LookupError:
    return []
else:
    extension = app_config.cms_extension
    return extension.navigation_apps_models
```

djangocms_navigation/utils.py
```diff
@@ -0,0 +1,13 @@
+from functools import lru_cache
+
+from django.apps import apps
+
+
+@lru_cache(maxsize=1)
+def supported_models():
```
monikasulik: Missing a test for this.

monikasulik:
@webbyfox Although you've added validation for not allowing adding of menu items in the "root" position, it's still possible to add root menu items by adding them as siblings to existing roots (just managed to do this). So I think this is a bug.

PR #9, opened 2019-01-04
Model changes for versioning integration

monikasulik:


djangocms_navigation/models.py
```diff
@@ -27,20 +27,26 @@ class Meta:
 
 
 class MenuContent(models.Model):
-    title = models.CharField(verbose_name=_("title"), max_length=100)
     menu = models.ForeignKey(Menu, on_delete=models.PROTECT)
+    root = models.ForeignKey('djangocms_navigation.MenuItem')
```
webbyfox: We should use a class on Foreign key ref rather than string and on_delete we should protect menucontent entry. Also, we should allow `null=true` as in most cases user will add menucontent entry first and then menuitem so user will come back to save root.

djangocms_navigation/test_utils/factories.py
```diff
@@ -0,0 +1,142 @@
...
+
+from django.contrib.auth.models import User
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.sites.models import Site
+
+from cms.models import Page, PageContent, TreeNode
+
+from djangocms_versioning.models import Version
+
+import factory
```
webbyfox: isort might fail here.

djangocms_navigation/test_utils/factories.py
```diff
@@ -0,0 +1,142 @@
...
+
+class TreeNodeFactory(factory.django.DjangoModelFactory):
+    site = factory.fuzzy.FuzzyChoice(Site.objects.all())
+    depth = 0
+    # NOTE: Generating path this way is probably not a good way of
+    # doing it, but seems to work for our present tests which only
+    # really need a tree node to exist and not throw unique constraint
+    # errors on this field. If the data in this model starts mattering
+    # in our tests then something more will need to be done here.
+    path = FuzzyText(length=8, chars=string.digits)
```
webbyfox: A bit off context comment here but we might need to think about how would we deal to create a version when path has alphabet like this `000E`?

djangocms_navigation/models.py
monikasulik: I think we should create the MenuItem object automatically. @chaosk Would you rather this was done on the model level (i.e. on MenuContent save) or on the admin level? If on the model level I can add it to this PR, otherwise it'll be in mine or @webbyfox's next PR I guess.

djangocms_navigation/test_utils/factories.py
monikasulik: I think it does it alphabetically, no? Besides, I think this is copy and pasted from other factories.py files, so wouldn't that mean it fails everywhere?

djangocms_navigation/test_utils/factories.py
monikasulik: Yeah... True. The way I see it, I'll figure that out once the problem arises in our tests ;) So far this works, but yeah - we'll probably need a better way of generating the path eventually.

djangocms_navigation/models.py
chaosk: Think model's save is better here.

monikasulik: @webbyfox So actually, defining foreign keys with strings is pretty standard: https://docs.djangoproject.com/en/2.1/ref/models/fields/#foreignkey It's also in some cases a good way to avoid circular imports and stuff like that. I can change it from a string to the class, but then I'll have to reorder the model classes and I actually thought they made more sense in the present order (as Menu is the "parent" of MenuContent, MenuContent of MenuItem).

djangocms_navigation/test_utils/factories.py
webbyfox: I thought all import should be a group on top and alphabetically sort

monikasulik: Just tried running isort and it did something that looked ridiculous and made no sense to me :D (it also complained about 4 other files in the repo). I can commit the isort changed version if you prefer or I can just do what pep8 recommendation suggest (i.e. three import groups: https://www.python.org/dev/peps/pep-0008/#imports) or do nothing :P Please choose :D

webbyfox: pep8 recommendation, please if you don't mind.

webbyfox: not as expected but I will leave it for now and approve.

PR #10, opened 2019-01-08
WIP: Versioning navigation

monikasulik:


djangocms_navigation/admin.py
```diff
@@ -128,6 +136,13 @@ def change_view(
         extra_context = extra_context or {}
         if menu_content_id:
             request.menu_content_id = menu_content_id
+            if self._versioning_enabled:
+                menu_content = get_object_or_404(
+                    MenuContent._base_manager, id=menu_content_id)
+                version = Version.objects.get_for_content(menu_content)
+                if version.state != DRAFT:
```
chaosk: What do you think about the following:
```
from djangocms_versioning.conditions import ConditionFailed
try:
    version.check_modify()
except ConditionFailed as e:
    messages.error(request, e)
```

This takes care of checking other conditions as they appear, for example moderation checks if the object is already in a collection to fail the check.

djangocms_navigation/cms_config.py
```diff
@@ -23,9 +28,54 @@ def configure_app(self, cms_config):
...
+
+    # Copy menu items
+    for item in MenuItem.get_tree(original_root).exclude(pk=original_root.pk):
+        item_fields = {
+            field.name: getattr(item, field.name)
+            for field in MenuItem._meta.fields
+            if field.name not in [MenuItem._meta.pk.name, "path"]
+        }
+        item_fields["path"] = new_root.path + item.path[MenuItem.steplen:]
+        MenuItem.objects.create(**item_fields)
```
chaosk: Can we use bulk_create here? Should improve performance.

djangocms_navigation/cms_config.py
```diff
@@ -23,9 +28,54 @@ def configure_app(self, cms_config):
...
+
+    return new_content
+
+
+class NavigationCMSAppConfig(CMSAppConfig):
+    djangocms_navigation_enabled = getattr(
+        settings, "NAVIGATION_CMS_MODELS_ENABLED", True
```
chaosk: I'd suggest prefixing the settings with DJANGOCMS_NAVIGATION

djangocms_navigation/cms_config.py
monikasulik: @chaosk I'm not sure... treebeard does a lot of funky custom db stuff and and django's `bulk_create`  also has some caveats such as not sending out signals or not running the model's save method etc. So I'm a little concerned something might go wrong with `bulk_create`. But if you prefer to go that route anyway, I can.

djangocms_navigation/admin.py
```diff
@@ -177,10 +200,21 @@ def response_add(self, request, obj, post_url_continue=None):
         return HttpResponseRedirect(url)
 
     def move_node(self, request, menu_content_id):
+        # Disallow moving of a node on anything other than a draft version
+        if self._versioning_enabled:
+            menu_content = get_object_or_404(
+                MenuContent._base_manager, id=menu_content_id)
+            version = Version.objects.get_for_content(menu_content)
+            if version.state != DRAFT:
+                # TODO: Should this be a 400 rather than 403?
```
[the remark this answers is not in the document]
chaosk: Well, it doesn't seem like a permission problem to me, so 400 makes a bit more sense.

400 below is just returning plain string, either adapt that to a JSON, or make this like return plain string as well. Just make it consistent.

djangocms_navigation/cms_config.py
chaosk: I've checked treebeard's code and `bulk_create` should work in this case. I'd still recommend checking if it doesn't break anything.

monikasulik: @chaosk I've got a unit test already where it creates about 7 variously nested menu items. Not really sure what else I can do to test this more thoroughly, is that sufficient do you think?

chaosk: Should be fine then!

PR #18, opened 2019-01-15
Improvement for Menu Item form using Select2 js lib

webbyfox:


djangocms_navigation/cms_config.py
```diff
@@ -8,13 +8,13 @@
 
 class NavigationCMSExtension(CMSAppExtension):
     def __init__(self):
-        self.navigation_apps_models = []
+        self.navigation_apps_models = {}
 
     def configure_app(self, cms_config):
         if hasattr(cms_config, "navigation_models"):
             navigation_app_models = getattr(cms_config, "navigation_models")
             if isinstance(navigation_app_models, Iterable):
```
chaosk: Gotta be a Mapping now, right?

djangocms_navigation/forms.py
```diff
@@ -47,17 +69,6 @@ def __init__(self, *args, **kwargs):
...
-            content_choices.extend([(obj.id, obj) for obj in model.objects.all()])
-        self.fields["object_id"].widget.choices = content_choices
-
-        # TODO: If this initial still needed after autocomplete changes
-        # then don't forget to write tests for this
-        if self.instance:
-            self.fields["object_id"].initial = self.instance.object_id
-
     def clean(self):
         cleaned_data = super().clean()
```
chaosk: Perhaps add validation for content_type/object_id, whether there's actually an object in the db.

djangocms_navigation/test_utils/polls/utils.py
```diff
@@ -0,0 +1,6 @@
+def get_all_poll_content_objects(model, **kwargs):
+    return model.objects.all()
+
+
+def get_published_pages_objects(model, site, **kwargs):
```
chaosk: what's this used for?

djangocms_navigation/forms.py
webbyfox: This could be tricky as content_type and object_id is not necessary for root item.

djangocms_navigation/test_utils/polls/utils.py
webbyfox: removed. copy/paste :smiley_cat:

djangocms_navigation/forms.py
chaosk: well, detect it's a root item then and skip validation

djangocms_navigation/views.py
```diff
@@ -0,0 +1,91 @@
...
+from django.core.exceptions import PermissionDenied
+from django.http import JsonResponse
+from django.views.generic import ListView
+
+from cms.models import Page
+
+from djangocms_navigation.utils import is_model_supported, supported_models
+
+
+class ContentObjectSelect2View(ListView):
```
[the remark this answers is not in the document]
webbyfox: We expecting multiple results also we need to filter queryset based on a user query provided in UI which makes more sense to use `ListView` rather just `View`. As `ListView` has `get_queryset` method to override which I don't think `View` has.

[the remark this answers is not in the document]
monikasulik: Yeah, but then you're overriding `get()` without using `super` and overriding `get_queryset` without `super` and not even setting any of the attributes that a `ListView` would expect (cause you've overridden everything that would error out without those attrs), so really you may as well make this officially a custom view by inheriting from `View` I think, no? Otherwise if somebody looks just at the class definition of this view (rather than the overridden methods), they might have an incorrect assessment of what this view is.

djangocms_navigation/views.py
```diff
@@ -0,0 +1,91 @@
...
+    def get(self, request, *args, **kwargs):
+
+        self.site = get_current_site(request)
+        self.object_list = self.get_queryset()
+        context = self.get_context_data()
+        data = {
+            "results": [
+                {"text": str(obj), "id": obj.pk} for obj in context["object_list"]
```
monikasulik: Why aren't you just doing `for obj in self.get_queryset()` btw? Since you're not returning a template response (but a json response), you don't actually need context data for anything AFAIK?

webbyfox: Agreed. I changed method name to get_data to make it clear and also used View as base class.

webbyfox:
@monikasulik Select2 js lib is part of core cms hence used here.

README.rst
```diff
@@ -26,3 +26,22 @@ Run::
 
 to perform the application's database migrations.
 
+App Integration
+===============
+
+To register model to use navigation app, app should provide class in cms_config.py which inherit `CMSAppConfig`
+class. It should have `djangocms_navigation_enable` flag True which register to use djangocms_navigation and
```
monikasulik: typo: djangocms_navigation_enable**d**

README.rst
```diff
@@ -26,3 +26,22 @@ Run::
...
+provide model mapping object, `navigation_models`.
+
+Mapping object should provide Model class as key and list of model fields which can be query from UI. Example of
```
monikasulik: I'd change this to "list of model fields which will be used for autocomplete form fields" as "query from UI" isn't too clear.

djangocms_navigation/forms.py
```diff
@@ -23,8 +26,32 @@ class Meta:
...
+        js = ("cms/js/select2/select2.js", "djangocms_navigation/js/create_url.js")
+
+
+class ContentTypeObjectSelectWidget(Select2Mixin, forms.TextInput):
+    def get_url(self):
+        return admin_reverse(SELECT2_CONTENT_OBJECT_URL_NAME)
+
+    def build_attrs(self, *args, **kwargs):
+        attrs = super().build_attrs(*args, **kwargs)
+        # import pdb; pdb.set_trace()
```
monikasulik: left over from debugging

djangocms_navigation/views.py
```diff
@@ -2,25 +2,38 @@
...
+        if not content_type_id:
+            return HttpResponseBadRequest()
+
+        # return http bad request if content_type not exist in db
+        try:
+            content_object = ContentType.objects.get_for_id(content_type_id)
+        except ContentType.DoesNotExist:
+            return HttpResponseBadRequest()
+
+        # return http request if content type is not registered to use navigation app
```
monikasulik: typo: you skipped the `bad` bit ;) (as in `http bad request`)

djangocms_navigation/views.py
```diff
@@ -2,25 +2,38 @@
...
+        model = content_object.model_class()
+        if not is_model_supported(model):
+            return HttpResponseBadRequest()
+
         self.site = get_current_site(request)
-        self.object_list = self.get_queryset()
-        context = self.get_context_data()
+        self.object_list = self.get_data()
```
monikasulik: don't think you need to assign this to a class attribute?

djangocms_navigation/views.py
```diff
@@ -29,24 +42,17 @@ def dispatch(self, request, *args, **kwargs):
...
         query = self.request.GET.get("query", None)
         site = self.request.GET.get("site")
+
         try:
             content_object = ContentType.objects.get_for_id(content_type_id)
         except ContentType.DoesNotExist:
-            raise ValueError(
-                "Content type with id {} does not exists.".format(content_type_id)
-            )
+            return HttpResponseBadRequest()
```
monikasulik: Aren't you already doing all these checks in the `get` method now? Would be good to keep all these validation checks in one place rather than spread across two methods.

tests/test_views.py
```diff
@@ -2,26 +2,36 @@
...
 
     def test_select2_view_endpoint_post(self):
         """HTTP post shouldn't allowed on this endpoint"""
         response = self.client.post(self.select2_endpoint)
-        self.assertEqual(response.status_code, 403)
+        expected_url = "/en/admin/login/?next=" + self.select2_endpoint
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response.url, expected_url)
+
+    def test_select2_view_endpoint_user_with_no_perm(self):
```
monikasulik: I think there probably should also be a test for a post request and user with no perms.

tests/test_views.py
```diff
@@ -89,9 +101,11 @@ def test_select2_view_search_exact_text_page(self):
             )
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.json()["results"]), 1)
+        # our query should be in text of resultset
+        self.assertIn("test2", response.json()["results"][0]["text"])
 
     def test_select2_view_dummy_search_text_page(self):
-        """ Both page should appear in results for dummy query"""
+        """ query which doesnt match shoul return 0 results"""
```
monikasulik: typo: should

tests/test_views.py
```diff
@@ -107,7 +121,11 @@ def test_select2_view_text_poll_content_repr(self):
                 data={"content_type_id": self.poll_content_contenttype_id},
             )
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.json()["results"][0]["text"], str(self.poll_content))
+        expected_result_list = [
+            {"text": "example", "id": 1},
+            {"text": "example2", "id": 2},
+        ]
+        self.assertListEqual(response.json()["results"], expected_result_list)
```
monikasulik: So lets say the json looks like this:
```
{
    'results': [...],
    'blablabla': 'cats',
}
```
Is it correct for the test to still pass? (it would) I would have thought that no.

tests/utils.py
```diff
@@ -9,3 +11,7 @@ def assertNotRaises(self, exc_type):
             yield None
         except exc_type:
             raise self.failureException("{} raised".format(exc_type.__name__))
+
+
+class WidgetTestForm(forms.Form):
```
monikasulik: I found this quite strange since it says WidgetTestForm and then doesn't contain a widget! Consider whether you actually need this in test/utils.py IMO you could easily just define this class inside of the test as you're only using this form in one test and are unlikely to use it elsewhere.

djangocms_navigation/views.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self.site = get_current_site(request)
+        self.object_list = self.get_data()
+
+        data = {
+            "results": [{"text": str(obj), "id": obj.pk} for obj in self.get_data()]
+        }
+        return JsonResponse(data)
+
+    def dispatch(self, request, *args, **kwargs):
```
monikasulik: Can't the whole `dispatch` override be removed now that the admin site decorator has been added to `get_urls`?

tests/base.py
```diff
@@ -0,0 +1,62 @@
...
+        self.poll_content2 = PollContent.objects.create(
+            poll=self.poll, language=self.language, text="example2"
+        )
+        self.page_contenttype_id = ContentType.objects.get_for_model(Page).id
+        self.poll_content_contenttype_id = ContentType.objects.get_for_model(
+            PollContent
+        ).id
+
+    @classmethod
+    def is_versioning_enabled(cls):
```
monikasulik: I don't think this method is being used anymore.

tests/base.py
```diff
@@ -0,0 +1,62 @@
...
+        return "djangocms_versioning" in settings.INSTALLED_APPS
+
+    def _get_version(self, grouper, version_state, language=None):
+        language = language or self.language
+
+        from djangocms_versioning.models import Version
```
monikasulik: Why is the import here and not at the top of the file?

tests/base.py
```diff
@@ -0,0 +1,62 @@
...
+from djangocms_navigation.test_utils.factories import PageContentFactory
+from djangocms_navigation.test_utils.polls.models import Poll, PollContent
+
+
+class BaseViewTestCase(CMSTestCase):
+    """setup data  for view test case"""
+
+    select2_endpoint = admin_reverse(SELECT2_CONTENT_OBJECT_URL_NAME)
+
+    def setUp(self):
```
monikasulik: I still feel this entire class should be deleted as it's just a way to tempt people to create a lot more db objects in their tests than they actually need (how many times will we need 2 users, 2 pages, 2 sites and 2 polls all in one test? do you even need all these objects at the same time in any of your tests?). And these sorts of things do tend to have consequences later on (like tests becoming excruciatingly slow and then people finding them too annoying to regularly run them as a result). Currently, there's simply no reason you can't just define what you need directly in the test case you're using this in.
But if, for some reason, you insist on keeping this class then at least change `setUp` to `setUpTestData` which will run once per TestCase rather than once per each test in the TestCase (as `setUp` does). That will help with not slowing down the tests as much.

djangocms_navigation/forms.py
```diff
@@ -82,6 +92,24 @@ def clean(self):
...
+            and cleaned_data["object_id"]
+        ):
+            try:
+                obj = (
+                    cleaned_data["content_type"]
+                    .model_class()
+                    .objects.get(pk=cleaned_data["object_id"])
+                )
+            except cleaned_data["content_type"].model_class().DoesNotExist:
+                obj = None
```
chaosk: Can just raise the validation error here

djangocms_navigation/views.py
```diff
@@ -0,0 +1,86 @@
...
+
+        if pk:
+            queryset = queryset.filter(pk=pk)
+
+        if query:
+            # For Page model loop through all title objects to exclude the
+            # object which doesnt match query
+            if model == Page:
+                for item in queryset:
+                    if item.get_page_title().lower().find(query.lower()) == -1:
```
chaosk: I think this can be improved in the future. You could filter via the title_set somehow, instead of doing the filtering in python

PR #22, opened 2019-01-28
Added preview URL for MenuItem

webbyfox:


tests/test_views.py
```diff
@@ -1,13 +1,85 @@
...
+
+    def test_view_standard_user(self):
+        standard_user = self.get_standard_user()
+        with self.login_user_context(standard_user):
+            response = self.client.get(self.preview_url)
+            expected_url = "/en/admin/login/?next=" + self.preview_url
+            self.assertEqual(response.status_code, 302)
+            self.assertEqual(response.url, expected_url)
+
+    def test_view_super_user(self):
```
monikasulik: I would suggest removing this test (if you follow the comments below, this test will not be needed) and perhaps dividing these tests into two TestCase classes. One which is just for testing permissions and the other for everything else to do with this view. This has two advantages. Firstly, you can put the creation of the superuser and log the user in in the `setUp` method in the "everything else TestCase" (which you can't in the permission test case since it's a different user in every test). Secondly, it allows more granular running of tests during development (sometimes it's useful to be able to run all tests in a smaller TestCase).

tests/test_views.py
```diff
@@ -1,13 +1,85 @@
...
+            self.assertEqual(response.status_code, 200)
+
+    def test_view_context_data_with_valid_menu_content(self):
+        factory = RequestFactory()
+        request = factory.get(self.preview_url)
+        request.user = self.get_superuser()
+        view = MenuContentPreviewView()
+        view.request = request
+        view.kwargs = {"menu_content_id": self.menu_content.id}
+        response = view.get_context_data()
```
monikasulik: This isn't a response actually, just context data.

tests/test_views.py
```diff
@@ -1,13 +1,85 @@
...
+
+    def test_view_context_data_with_invalid_int_menu_content(self):
+        factory = RequestFactory()
+        request = factory.get(self.preview_url)
+        request.user = self.get_superuser()
+        view = MenuContentPreviewView()
+        view.request = request
+        view.kwargs = {"menu_content_id": 99}
+        response = view.get_context_data()
+        self.assertEqual(response.status_code, 400)
```
monikasulik: IMO this test should definitely use `self.client.get` to get the response. Usually `get_context_data` returns a dict, not an http response, so if for some reason the 400 response doesn't get raised, this test is going to come up with a confusing and unhelpful exception - an AttributeError that says a dict doesn't have a status_code attribute. If you do this test by making an http call with `self.client.get` then you can reliably expect to get an HttpResponse object no matter what data you provide and get more helpful test output like 200 != 400 for example.

tests/test_views.py
```diff
@@ -1,13 +1,85 @@
...
+
+    def test_view_context_data_with_invalid_string_menu_content(self):
+        factory = RequestFactory()
+        request = factory.get(self.preview_url)
+        request.user = self.get_superuser()
+        view = MenuContentPreviewView()
+        view.request = request
+        view.kwargs = {"menu_content_id": "dummy"}
+        response = view.get_context_data()
+        self.assertEqual(response.status_code, 400)
```
monikasulik: Same as above - really should be using `self.client.get`

tests/test_views.py
```diff
@@ -1,85 +1,82 @@
...
+
+class PreviewViewTestCases(CMSTestCase):
+    def setUp(self):
+        self.menu_content = MenuContentFactory()
+        self.preview_url = admin_reverse(
+            "djangocms_navigation_menuitem_preview",
+            kwargs={"menu_content_id": self.menu_content.id},
+        )
+
+    def test_view_url(self):
```
monikasulik: As per previous comments - I think this test can be removed, unless you have a reason why you think checking the url pattern like this gives us some sort of safeguard?

tests/test_views.py
```diff
@@ -1,85 +1,82 @@
...
-        factory = RequestFactory()
-        request = factory.get(self.preview_url)
-        request.user = self.get_superuser()
-        view = MenuContentPreviewView()
-        view.request = request
-        view.kwargs = {"menu_content_id": "dummy"}
-        response = view.get_context_data()
-        self.assertEqual(response.status_code, 400)
+        super_user = self.get_superuser()
+        with self.login_user_context(super_user):
```
monikasulik: Now that you've divided this up into two test cases, you can log in the super user in `setUp`. Saves you having to write these two lines in every test.

tests/test_views.py
```diff
@@ -1,85 +1,82 @@
...
+    def test_view_response_with_invalid_int_menu_content(self):
+        super_user = self.get_superuser()
+        with self.login_user_context(super_user):
+            preview_url = admin_reverse(
+                "djangocms_navigation_menuitem_preview", kwargs={"menu_content_id": 99}
+            )
+            response = self.client.get(preview_url)
+            self.assertEqual(response.status_code, 404)
+
+    def test_view_url_with_invalid_string_menu_content(self):
```
monikasulik: I do think this test is probably slightly overkill. I mean testing for the non-existence of urls means we could potentially be testing this for any url! It made sense to have a test for ValueError, but not sure it does for this.

tests/test_views.py
```diff
@@ -1,85 +1,82 @@
...
 
-    def test_view_super_user(self):
-        staff_user = self.get_superuser()
-        with self.login_user_context(staff_user):
-            response = self.client.get(self.preview_url)
-            self.assertEqual(response.status_code, 200)
+
+class PreviewViewTestCases(CMSTestCase):
+    def setUp(self):
+        self.menu_content = MenuContentFactory()
```
monikasulik: Correct me if I'm wrong but I think you're only using this db object in just one test in this test case, so makes more sense to define it there rather than in `setUp`.

PR #23, opened 2019-01-29
Selecting different menus in navigation plugin

monikasulik:


djangocms_navigation/models.py
```diff
@@ -64,6 +70,7 @@ class NavigationPlugin(CMSPlugin):
         default=TEMPLATE_DEFAULT,
         max_length=255,
     )
+    menu = models.ForeignKey(Menu)
```
chaosk: Needs a `on_delete`

PR #25, opened 2019-02-04
Added validation for MenuItem form

webbyfox:


djangocms_navigation/forms.py
```diff
@@ -74,6 +74,10 @@ def clean(self):
         if self.errors:
             return cleaned_data
 
+        # Setting object_id None if user has not selected as will be empty string which will raise ValueError.
```
chaosk: Rewrite this a bit:
```object_id = cleaned_data.get("object_id")
content_type = cleaned_data.get("content_type")
```

And then use `object_id` and `content_type` instead of `cleaned_data.get("FOO")` and `cleaned_data["FOO"]`

PR #26, opened 2019-02-11
Fil 846 integrate version locking

geoidesic:


djangocms_navigation/cms_config.py
```diff
@@ -82,3 +82,8 @@ class NavigationCMSAppConfig(CMSAppConfig):
             preview_url=MenuContent.get_preview_url,
         )
     ]
+
+    djangocms_versionlocking_enabled = getattr(
+        settings, 'VERSION_LOCKING_CMS_MODELS_ENABLED', True
```
chaosk: This setting should be called `DJANGOCMS_NAVIGATION_VERSION_LOCKING_ENABLED`

tests/test_appregistration.py
```diff
@@ -232,3 +232,50 @@ def test_navigation_is_moderated_if_moderation_setting_disabled(self):
...
+    def tearDown(self):
+        """Populate everything again so our setting changes do not
+        effect any other tests"""
+        # Set the defaults for the navigation app config again
+        imp.reload(cms_config)
+        navigation_app = apps.get_app_config('djangocms_navigation')
+        navigation_app.cms_config = cms_config.NavigationCMSAppConfig(navigation_app)
+        # Reset the versioning app
+        self.versionlocking_app.cms_extension.version_lock_models = []
+        configure_cms_apps([self.versionlocking_app])
```
webbyfox: We found leaking while running tests yesterday and Monika found the cause and she pushes the fix on my PR. I think we need the same fix here as well. see https://github.com/FidelityInternational/djangocms-navigation/pull/33/files#diff-fe0ef300e68c61297ab3d98a8f80b0f2R143

geoidesic: I'll need a bit more info please – there's a lot to look at on that link you sent. Could you describe the problem so I can recognise which fix you're referring to?

webbyfox: In various setting test cases we manipulate various app configurations and while triggering teardown method we were not setting those configurations back to it was before the test which was the cause for other related test behave unexpectedly. Monika fixes it I sent the link with line number which needed to add in above teardown method.

webbyfox:
Closing this PR as FIL 900 story will remove app registration from version locking. FYI, Version locking is creating a lock for all model that registered with versioning.

PR #33, opened 2019-02-12
Fix for multiple versions

webbyfox:


tests/test_cms_menus.py
```diff
@@ -0,0 +1,104 @@
...
+        self.assertEqual(roots.count(), 2)
+
+    @disable_versioning_for_navigation()
+    def test_get_roots_multiple_version_no_versioning(self):
+        root_1 = factories.RootMenuItemFactory()
+        root_1_version_1 = factories.MenuVersionFactory(content__root=root_1)
+        root_1_version_1.archive(self.user)
+        root_1_version_2 = root_1_version_1.copy(self.user)
+        root_2 = factories.RootMenuItemFactory()
+        root_2_version_1 = factories.MenuVersionFactory(content__root=root_2)
```
monikasulik: Btw, you're obviously going to need to use the factories that don't create versions for a test that checks behaviour with versioning disabled :)

tests/test_cms_menus.py
```diff
@@ -0,0 +1,104 @@
...
+            title=child2.title,
+            url=child2.content.get_absolute_url(),
+            id=child2.id,
+            parent_id=menu_contents[1].menu.root_id,
+            attr={"link_target": child2.link_target},
+        )
+
+    def test_get_roots_with_single_version(self):
+        menucontent_1 = factories.MenuContentFactory()
+        menucontent_2 = factories.MenuContentFactory()
```
monikasulik: This actually checks how it behaves with 0 versions rather than 1 version.

djangocms_navigation/cms_menus.py
```diff
@@ -15,8 +15,9 @@ def get_roots(self, request):
         queryset = MenuItem.get_root_nodes().filter(
             menucontent__menu__site=get_current_site()
         )
-        menucontents = get_versionable_for_content(MenuContent).distinct_groupers()
+        menucontents = get_versionable_for_content(MenuContent)
```
chaosk: Rename to versionable

djangocms_navigation/utils.py
```diff
@@ -37,4 +37,4 @@ def get_versionable_for_content(content):
     try:
         return versionables.for_content(content)
     except KeyError:
-        pass
+        return
```
chaosk: What's the difference?

webbyfox: IMO, `pass` is to create an empty block where the return is to let control flow to continue which is more readable I think ;)

djangocms_navigation/cms_menus.py
```diff
@@ -5,15 +5,21 @@
...
 from menus.menu_pool import menu_pool
 
-from .models import MenuItem
+from .models import MenuContent, MenuItem
+from .utils import get_versionable_for_content
 
 
 class CMSMenu(Menu):
     def get_roots(self, request):
         # TODO: What if the MenuItem objects are related to an unpublished version?
```
monikasulik: Now that this method is versioning aware, is this comment also resolved? Is this an issue at all? Should the tests test for this case more specifically - as in create multiple versions with multiple states? @chaosk Do you have any opinions?

webbyfox: I think the method is version aware but todo is still relevant as let say published nav is unpublished then out method will return unpublish version of menu content as it was the latest version of it which still the case to handle? @chaosk

tests/test_cms_menus.py
```diff
@@ -75,37 +74,61 @@ def test_get_nodes(self):
...
-
-    def test_get_roots_with_versioning_enabled(self):
+    def test_get_roots_with_versions(self):
         """This test to check versioning would group all the versions
         of menu content and return latest of all distinct menu content
         """
+        menuitem_1 = factories.RootMenuItemFactory()
+        menucontent_1_version_1 = factories.MenuVersionFactory(content__root=menuitem_1)
+        menucontent_1_version_1.archive(self.user)
+        menucontent_1_version_2 = menucontent_1_version_1.copy(self.user)
```
monikasulik: You shouldn't have to run the `copy` method as part of data setup for a test. It's a complex method, which could have all kinds of side effects and coupling test data setup with it really isn't a good idea. `MenuVersionFactory(content__menu=menucontent_1_version1.content.menu)` should give you the same data as `copy`.

tests/test_cms_menus.py
```diff
@@ -75,37 +74,61 @@ def test_get_nodes(self):
...
+        menuitem_2 = factories.RootMenuItemFactory()
```
monikasulik: I thought the reason you created the root menu items separately was to be able to assert them more easily below, but then you're not using those vars, so what do you need them for?

tests/test_cms_menus.py
```diff
@@ -75,37 +74,61 @@ def test_get_nodes(self):
...
+            list(roots),
+            [
+                menucontent_1_version_2.content.root,
+                menucontent_2_version_1.content.root,
+            ],
+        )
 
     @disable_versioning_for_navigation()
-    def test_get_roots_with_versioning_disabled(self):
+    def test_get_roots_with_when_versions_in_db_while_versioning_disable(self):
```
monikasulik: typo: disabled rather than disable

tests/test_cms_menus.py
```diff
@@ -75,37 +74,61 @@ def test_get_nodes(self):
...
+        menucontent_3 = factories.MenuContentFactory()
+        roots = self.menu.get_roots(self.request)
+        self.assertEqual(roots.count(), 3)
+        self.assertListEqual(
+            list(roots), [menucontent_1.root, menucontent_2.root, menucontent_3.root]
+        )
+
+
+@disable_versioning_for_navigation()
+class CMSMenuDisabledVersioningTestCase(CMSTestCase):
```
monikasulik: Why do you need this test case? What are you testing here that you aren't already in the test case above?

tests/test_cms_menus.py
```diff
@@ -0,0 +1,109 @@
...
+        self.assertNavigationNodeEqual(
+            nodes[4],
+            title=child2.title,
+            url=child2.content.get_absolute_url(),
+            id=child2.id,
+            parent_id=menu_contents[1].menu.root_id,
+            attr={"link_target": child2.link_target},
+        )
+
+    def test_get_roots_with_multiple_menucontents(self):
```
monikasulik: So I'm not really following why this test is still here? IMO `test_get_roots_with_versions` and `test_get_roots_with_versioning_disabled` already cover what's needed, no? Is there something this test gives you that the others don't already check for?

webbyfox: My thinking was to check that navigation as expected where versioning is enabled and no version in db and check there is no side effect of versioning injected hence have a similar test case for versioning enabled/disable.

PR #34, opened 2019-02-18
Added navigation menu to toolbar, wrapped versioning import try:except and added URL to menuitem changelist

webbyfox:


djangocms_navigation/admin.py
```diff
@@ -269,6 +275,20 @@ def __new__(cls, *args, **kwargs):
 
         return Form
 
+    def get_list_display(self, request):
+        return list(super().get_list_display(request)) + ["get_object_url"]
```
chaosk: Just add to `list_display` instead of overriding `get_list_display`

djangocms_navigation/admin.py
```diff
@@ -269,6 +275,20 @@ def __new__(cls, *args, **kwargs):
...
+
+    def get_object_url(self, obj=None):
```
chaosk: Don't think this can ever be `None`

djangocms_navigation/admin.py
```diff
@@ -269,6 +275,20 @@ def __new__(cls, *args, **kwargs):
...
+        if obj is None:
+            return
+        if obj.content_type and obj.object_id:
+            ct = ContentType.objects.get_for_id(obj.content_type.id)
```
chaosk: `ct` is `obj.content_type`, that's unnecessary

djangocms_navigation/admin.py
```diff
@@ -269,6 +275,20 @@ def __new__(cls, *args, **kwargs):
...
+            object = ct.get_object_for_this_type(pk=obj.object_id)
```
chaosk: `obj.content` instead of all this

djangocms_navigation/admin.py
```diff
@@ -269,6 +275,20 @@ def __new__(cls, *args, **kwargs):
...
+            obj_url = object.get_absolute_url()
+            return format_html('<a href="' + obj_url + '">' + obj_url + "</a>")
+
+    get_object_url.short_description = "URL"
```
chaosk: `_()`

djangocms_navigation/admin.py
```diff
@@ -24,6 +20,15 @@
 from .views import ContentObjectSelect2View, MenuContentPreviewView
 
 
+# TODO: Tests to be added (FIL-908).
```
geoidesic: We can't really include FIL numbers when submitting to Divio repo

geoidesic: FIL number is supposed to be in the FIL pull request (but not the divio pull request)... makes it hard for me to know which story this PR relates to.

djangocms_navigation/admin.py
```diff
@@ -269,6 +275,13 @@ def __new__(cls, *args, **kwargs):
 
         return Form
 
+    def get_object_url(self, obj):
+        if obj.content:
+            obj_url = obj.content.get_absolute_url()
+            return format_html('<a href="' + obj_url + '">' + obj_url + "</a>")
```
chaosk: This is not a safe way of building HTML, use: https://docs.djangoproject.com/en/2.1/ref/utils/#django.utils.html.format_html

djangocms_navigation/admin.py
webbyfox: If you double check, this branch has FIL number of the current story number. I've added FIL number on TODO task as work will be done as part of that story number. Navigation is Fidelity specific repo and not the Divio Repo. However, you still want me to remove it please let me know.

PR #38, opened 2019-03-07
Purge menu cache while creation/editing and various version state changes

webbyfox:


tests/test_plugin.py
```diff
@@ -13,55 +14,52 @@
 from .utils import disable_versioning_for_navigation
 
 
-class NavigationSelectorTestCase(TestCase):
+try:
+    from djangocms_versioning.constants import ARCHIVED, DRAFT, UNPUBLISHED, PUBLISHED
+except ImportError:
+    ARCHIVED, DRAFT, PUBLISHED, UNPUBLISHED = None
```
monikasulik: This doesn't look right to me... I don't think you should do try/except ImportError in tests as this could have unexpected consequences (like say a test might end up checking that something is equal to None instead of 'archived').
But also, I don't think you actually use this import anywhere?

tests/test_plugin.py
```diff
@@ -261,3 +269,268 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
+
+        # Getting response again to fetch child2 and grandchild2 menu item
+        response = self.client.get(page_url)
+        self.assertEqual(response.status_code, 200)
+
+        # rendering should generate cachekey object
+        cache_key = CacheKey.objects.all().count()
+        self.assertEqual(cache_key, 1)
+
+    def test_menu_cache_invalidate_after_menucontent_publish(self):
```
monikasulik: Why such complex tests for this? I don't understand why you need to create so much data (child-grandchild structures etc. which will be slower) when all you're asserting is the CacheKey? And of course the shorter a test the easier it is to read usually. So is there a reason for how you've chosen to do this?
I would think you achieve the same thing with something like:
```
def test_menu_cache_invalidated_on_publish(self):
    version = factories.MenuVersionFactory()
    self.assertEqual(CacheKey.objects.all().count(), 0)
    version.publish(self.get_superuser())
    self.assertEqual(CacheKey.objects.all().count(), 1)
```
Is that not so?

webbyfox: `CacheKey` only generated once the menu is rendered on the frontend. The test is creating `Menucontent` with child objects then rendering it on the frontend and then checking cache is invalidated or not then I add another child item and checking it appears as well.

tests/test_plugin.py
webbyfox: Sorry, this must be left while I was trying something...removed.

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
+
+        # Rendering should generate cachekey object
+        cache_key = CacheKey.objects.all().count()
+        self.assertEqual(cache_key, 1)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(child.title, str(response.content))
+        self.assertIn(grandchild.title, str(response.content))
+
+        # Create further menu items which covers draft state
```
monikasulik: What does "covers draft state" mean? And why does the draft state matter in this test? I thought this line was here simply because you want to check that when the menu version is modified, the cache clearing fires up?

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
 
         # Now delete the plugin from the page and assert
         self._delete_nav_plugin_and_assert(new_placeholder, new_plugin)
+
+    def test_menu_cache_invalidate_after_menuitem_creation(self):
```
monikasulik: I think this is actually a test for whether the cache is invalidated after edit rather than after creation, right?

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(child.title, str(response.content))
+        self.assertIn(grandchild.title, str(response.content))
+
+        # Create further menu items which covers draft state
+        child2 = factories.ChildMenuItemFactory(parent=menu_content.root)
+
+        # grandchild2
+        factories.ChildMenuItemFactory(parent=child2)
```
monikasulik: Tbh, I think you only need one of these two lines.

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
+        menu_content_version = factories.MenuVersionFactory()
+        menu_content = menu_content_version.content
+        child = factories.ChildMenuItemFactory(parent=menu_content.root)
+        # grandchild
+        factories.ChildMenuItemFactory(parent=child)
+
+        # Add nav plugin to placeholder
+        self._add_nav_plugin_and_assert(
+            placeholder, menu_content.menu, "menu/menu.html"
+        )
```
monikasulik: As above, ditto on TODO comment (and same for tests below)

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
+
+        # Now publish the page content containing the plugin,
+        # so the page can be viewed
+        version = page_content.versions.get()
+        version.publish(self.get_superuser())
```
monikasulik: As above, you probably won't need this if you just set this version to published straight away (and ditto for tests below).

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
+
+        # Making sure there is no cachekey existed before rendering page
+        cache_key = CacheKey.objects.all().count()
+        self.assertEqual(cache_key, 0)
+        # And view the page
+        page_url = page_content.page.get_absolute_url()
+        response = self.client.get(page_url)
+
+        # Rendering should generate cachekey object
```
monikasulik: This confused me, I thought you'd missed an intended assert, but then realized the comment actually describes an assert that is 4 lines down. So you might want to fix the line spacing/position of the comment/asserts here.

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
```
monikasulik: Ditto on strange line spacing/comment not describing the block of code it's attached to.

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
+
+        # Add nav plugin to placeholder
+        self._add_nav_plugin_and_assert(
+            placeholder, menu_content.menu, "menu/menu.html"
+        )
+
+        # Now publish the page content containing the plugin,
+        # so the page can be viewed
+        version = page_content.versions.get()
+        version.publish(self.get_superuser())
```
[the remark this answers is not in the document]
webbyfox: Not sure you can add a placeholder(and plugin) to the published page.

tests/test_plugin.py
```diff
@@ -261,3 +263,186 @@ def test_crud_for_navigation_plugin_when_versioning_disabled(self):
...
+
+        # And view the page
+        page_url = page_content.page.get_absolute_url()
+        response = self.client.get(page_url)
+
+        # Rendering should generate cachekey object
+        cache_key = CacheKey.objects.all().count()
+        self.assertEqual(cache_key, 1)
+
+        self.assertEqual(response.status_code, 200)
```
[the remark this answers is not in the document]
webbyfox: Not sure this assert needs any comment.

djangocms_navigation/utils.py
```diff
@@ -38,3 +38,9 @@ def get_versionable_for_content(content):
         return versionables.for_content(content)
     except KeyError:
         return
+
+
+def purge_menu_cache(site_id=None, language=None):
+    from menus.menu_pool import menu_pool
```
chaosk: Move this import to top level

tests/test_plugin.py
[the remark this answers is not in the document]
chaosk: I don't think you're modifying the page after this line

PR #41, opened 2019-03-26
Added filtering for get_roots based on version state

webbyfox:


djangocms_navigation/cms_menus.py
```diff
@@ -17,7 +19,8 @@ def get_roots(self, request):
         )
         versionable = get_versionable_for_content(MenuContent)
         if versionable:
-            menucontents = versionable.distinct_groupers()
+            inner_filter = {"versions__state__in": [DRAFT, PUBLISHED]}
```
monikasulik: Doesn't this mean that both DRAFT and PUBLISHED versions will always appear? And really what we want I guess is for both to appear in preview but only published ones to appear on the site, right?

tests/test_cms_menus.py
```diff
@@ -78,11 +78,13 @@ def test_get_nodes(self):
 
     def test_get_roots_with_multiple_menucontents(self):
         """Test to check get_roots while creating two menu contents"""
-        menucontent_1 = factories.MenuContentFactory()
-        menucontent_2 = factories.MenuContentFactory()
+        menucontent_1 = factories.MenuVersionFactory()
+        menucontent_2 = factories.MenuVersionFactory()
```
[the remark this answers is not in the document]
webbyfox: I am inclined to delete this test as there is already test as you mentioned does it for multiple version and there is a test further down (last one in file) with disabling versioning for nav.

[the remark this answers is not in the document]
monikasulik: I'd have to look at the tests in more detail, but if you're pretty sure that every case that needs to be covered is covered for both versioning enabled and disabled then I'm fine with deleting this one.

tests/test_cms_menus.py
```diff
@@ -76,32 +82,46 @@ def test_get_nodes(self):
...
-
-    def test_get_roots_with_versions(self):
+    def test_get_roots_when_draft_mode_isnt_active(self):
         """This test to check versioning would group all the versions
         of menu content and return latest of all distinct menu content
         """
         menucontent_1_v1 = factories.MenuVersionFactory(state=ARCHIVED)
+        # draft version
         menucontent_1_v2 = factories.MenuVersionFactory(
             content__menu=menucontent_1_v1.content.menu
```
monikasulik: As below, I think you're going to need a published version created here as well and perhaps also an unpublished one.

tests/test_cms_menus.py
```diff
@@ -76,32 +82,46 @@ def test_get_nodes(self):
...
+        menu = CMSMenu(renderer)
+
+        roots = menu.get_roots(self.request)
         self.assertEqual(roots.count(), 2)
         self.assertListEqual(
             list(roots), [menucontent_1_v2.content.root, menucontent_2_v1.content.root]
         )
 
-    def test_get_roots_with_archived_versions(self):
+    def test_get_roots_for_archived_and_unpublished_versions(self):
```
monikasulik: I think you need to test this for both draft mode active and inactive. But these don't have to be a separate tests. You could just add the creation of an unpublished version into the two tests above (and then this test can be deleted).

tests/test_cms_menus.py
```diff
@@ -82,33 +82,73 @@ def test_get_nodes(self):
             attr={"link_target": child2.link_target},
         )
 
-    def test_get_roots_when_draft_mode_isnt_active(self):
+    def test_get_nodes_for_versioning(self):
```
monikasulik: I think it's nicer if these tests are named `get_nodes_for_versioning_enabled` and `get_nodes_for_versioning_disabled`. Will make things clearer if one sees one of these fail in the output etc.

tests/test_cms_menus.py
```diff
@@ -82,33 +82,73 @@ def test_get_nodes(self):
...
-        menucontent_1_v1 = factories.MenuVersionFactory(state=ARCHIVED)
-        # draft version
-        menucontent_1_v2 = factories.MenuVersionFactory(
-            content__menu=menucontent_1_v1.content.menu
-        )
-        # draft version
-        menucontent_2_v1 = factories.MenuVersionFactory()
+        # Renderer should only render published menucontent
+        self.assertEqual(roots.count(), 1)
+        self.assertListEqual(list(roots), [menucontent_2_v1.content.root])
```
monikasulik: I think draft mode active is True is one test and draft mode active is False is another test (as it was before). The test I think you can remove now that you've added unpublished versions is `test_get_roots_for_archived_and_unpublished_versions`

tests/test_cms_menus.py
```diff
@@ -1,22 +1,27 @@
...
+from menus.menu_pool import menu_pool
 from .utils import disable_versioning_for_navigation
 
 
 try:
-    from djangocms_versioning.constants import ARCHIVED, DRAFT
+    from djangocms_versioning.constants import ARCHIVED, DRAFT, UNPUBLISHED, PUBLISHED
 except ImportError:
-    ARCHIVED, DRAFT = None
+    ARCHIVED, DRAFT, UNPUBLISHED = None
```
monikasulik: You should add PUBLISHED here as well

PR #50, opened 2019-09-03
Run tests through python setup.py test

jonathan-s:
This PR depends on 
- https://github.com/FidelityInternational/djangocms-version-locking/pull/55
- https://github.com/divio/djangocms-versioning/pull/189

To try this out locally see the two comments in code.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,15 @@ INSTALL_REQUIREMENTS = [
     "django-treebeard>=4.3",
 ]
 
+TEST_REQUIREMENTS = [
+    "djangocms_helper",
+    "djangocms_versioning",
+    "djangocms_version_locking",
+    "djangocms-moderation",
+    "factory_boy",
+    "django_cms"
+]
+
 
 setup(
     name="djangocms-navigation",
@@ -29,4 +38,11 @@ setup(
     url="https://github.com/fidelityInternational/djangocms-navigation",
     license="BSD",
     test_suite="tests.settings.run",
+    tests_require=TEST_REQUIREMENTS,
+    dependency_links=[
+        "http://github.com/divio/django-cms/tarball/release/4.0.x#egg=django-cms-4.0.0",
+        "http://github.com/divio/djangocms-versioning/tarball/master#egg=djangocms-versioning-0.0.23",
+        "http://github.com/divio/djangocms-moderation/tarball/release/1.0.x#egg=djangocms-moderation-1.0.x",
+        "http://github.com/FidelityInternational/djangocms-version-locking/tarball/master#egg=djangocms-version-locking-0.0.13",
+    ]
 )
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -2,3 +2,7 @@ djangocms-helper
 tox
 coverage
 isort
+factory_boy
+git+git://github.com/divio/djangocms-versioning.git#egg=djangocms-versioning
+git+git://github.com/FidelityInternational/djangocms-version-locking.git#egg=djangocms-version-locking
+git+git://github.com/divio/djangocms-moderation.git#egg=djangocms-moderation
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ local.sqlite
 htmlcov/
 .python-version
 !djangocms_versioning/static/djangocms_versioning/js/dist
+venv
--- a/README.rst
+++ b/README.rst
@@ -26,6 +26,21 @@ Run::
 
 to perform the application's database migrations.
 
+
+Running Tests
+=============
+
+You can run all the tests by executing:
+
+    python -m venv venv
+    source venv/bin/activate
+    python setup.py test
+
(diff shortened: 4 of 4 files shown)


setup.py
```diff
@@ -4,29 +4,44 @@
...
-    license="BSD",
-    test_suite="tests.settings.run",
+    author='Fidelity International',
+    url='https://github.com/fidelityInternational/djangocms-navigation',
+    license='BSD',
+    test_suite='tests.settings.run',
+    tests_require=TEST_REQUIREMENTS,
+    dependency_links=[
+        'http://github.com/divio/django-cms/tarball/release/4.0.x#egg=django-cms-4.0.0',
+        'http://github.com/divio/djangocms-versioning/tarball/master#egg=djangocms-versioning-0.0.23',
```
jonathan-s: change `divio` to `jonathan-s`

setup.py
```diff
@@ -4,29 +4,44 @@
...
+        'http://github.com/divio/djangocms-moderation/tarball/release/1.0.x#egg=djangocms-moderation-1.0.x',
+        'http://github.com/FidelityInternational/djangocms-version-locking/tarball/master#egg=djangocms-version-locking-0.0.13',
```
jonathan-s: change `FidelityInternational` to `jonathan-s`

setup.py
```diff
@@ -4,29 +4,44 @@
...
+    'djangocms_versioning',
+    'djangocms_version_locking',
+    'djangocms-moderation',
+    'factory_boy'
 ]
 
 
 setup(
-    name="djangocms-navigation",
+    name='djangocms-navigation',
```
Aiky30: This is good to go bar the changes to the quotes. Once this is reversed these changes can be merged.

jonathan-s:
I've made the quotes change now.

setup.py
```diff
@@ -29,4 +37,11 @@
     url="https://github.com/fidelityInternational/djangocms-navigation",
     license="BSD",
     test_suite="tests.settings.run",
+    tests_require=TEST_REQUIREMENTS,
+    dependency_links=[
+        "http://github.com/divio/django-cms/tarball/release/4.0.x#egg=django-cms-4.0.0",
+        "http://github.com/jonathan-s/djangocms-versioning/tarball/master#egg=djangocms-versioning-0.0.23",
+        "http://github.com/divio/djangocms-moderation/tarball/release/1.0.x#egg=djangocms-moderation-1.0.x",
+        "http://github.com/jonathan-s/djangocms-version-locking/tarball/master#egg=djangocms-version-locking-0.0.13",
```
Aiky30: This can't be jonathan-s

setup.py
```diff
@@ -29,4 +37,11 @@
     url="https://github.com/fidelityInternational/djangocms-navigation",
     license="BSD",
     test_suite="tests.settings.run",
+    tests_require=TEST_REQUIREMENTS,
+    dependency_links=[
+        "http://github.com/divio/django-cms/tarball/release/4.0.x#egg=django-cms-4.0.0",
+        "http://github.com/jonathan-s/djangocms-versioning/tarball/master#egg=djangocms-versioning-0.0.23",
```
Aiky30: This can't be jonathan-s

setup.py
```diff
@@ -5,10 +5,18 @@
 
 INSTALL_REQUIREMENTS = [
     "Django>=1.11,<2.0",
-    "django-cms>=3.5.0",
+    "django_cms",
```
Aiky30: Can this be left as:     "django-cms>=3.5.0", ?

setup.py
```diff
@@ -29,4 +37,11 @@
     url="https://github.com/fidelityInternational/djangocms-navigation",
     license="BSD",
     test_suite="tests.settings.run",
+    tests_require=TEST_REQUIREMENTS,
+    dependency_links=[
```
Aiky30: Are the dependancy links for tests only or will they also affect normal install of the package by downloading the packages? So in theory you could have the packages in .egg and the site-packages directory?

setup.py
jonathan-s: Bummer! My mistake.

setup.py
jonathan-s: The dependency links works for both tests and `install_requires`. I've changed the install require to use djangocms 3.5, when you run the tests it will install djangocms 4.0 instead. 

Does that answer your question?

setup.py
jonathan-s: Yes, this is now changed.

setup.py
Aiky30: It does, as long as that still works for the tests I would prefer to leave the install requirements as they were, if we changed that we would need to change the other addons too as they a have that requirement.

tests/requirements.txt
```diff
@@ -1,4 +1,7 @@
 djangocms-helper
 tox
 coverage
-isort
```
Aiky30: Did you intend to remove isort?

PR #51, opened 2019-09-04
Version locked items can't be moved or changed

jonathan-s:
Also refactors urs so that the admin can be used with other models that implements the same interface. 

Superseeded by https://github.com/FidelityInternational/djangocms-navigation/pull/56


jonathan-s:
Requires changes to some other files as well. So will be on that.

tests/test_admin.py
```diff
@@ -131,12 +132,63 @@ def test_get_queryset_filters_by_content_id(self):
...
+            "admin:djangocms_navigation_menuitem_move_node", args=(menu_content.id,)
+        )
+        self.data = {
+            "node_id": child_of_child.pk,
+            "sibling_id": menu_content.root.pk,
+            "as_child": 1,
+        }
+
+    def test_visit_change_view_when_node_is_version_locked_fails(self):
+        """It fails as super user is not the person who created the version"""
```
Aiky30: The docstrings should be on every test and should describes the business requirement with what and why. In this instance this could be: Visiting the change view for a menu item by a different user should be rejected, because the user is different to the locked to user.

This helps anyone making changes in the future who could break this test understand what it was trying to do, if it then needs to be reworked it can be as the developer should be informed of the requirement that the test is enforcing.

tests/test_admin.py
```diff
@@ -131,12 +132,63 @@ def test_get_queryset_filters_by_content_id(self):
...
+        response = self.client.get(self.change_url)
+        msg = list(get_messages(response.wsgi_request))[0]
+
+        self.assertEquals(msg.message, "You don't have permission to edit or it is locked")
+        self.assertEquals(response.status_code, 302)
+
+    def test_moving_node_that_is_version_locked_fails(self):
```
Aiky30: A menu item cannot be changed by a user who is not the lock owner.

tests/test_admin.py
```diff
@@ -131,12 +132,63 @@ def test_get_queryset_filters_by_content_id(self):
...
+        msg = "The item is currently locked or you don't have permission to change it"
+
+        self.assertEquals(response.status_code, 400)
+        self.assertEquals(content, msg)
+
+    @patch('djangocms_navigation.admin.using_version_lock', False)
+    def test_moving_node_version_lock_not_installed_works_without_error(self):
+        response = self.client.post(self.move_url, data=self.data)
+
+        self.assertEquals(response.status_code, 200)
```
Aiky30: We should really check that the page is that expected rather than just a 200 here.

tests/test_admin.py
```diff
@@ -131,12 +132,63 @@ def test_get_queryset_filters_by_content_id(self):
...
+    def test_moving_node_that_is_version_locked_fails(self):
+        response = self.client.post(self.move_url, data=self.data)
+        content = response.content.decode('utf-8')
+        msg = "The item is currently locked or you don't have permission to change it"
+
+        self.assertEquals(response.status_code, 400)
+        self.assertEquals(content, msg)
+
+    @patch('djangocms_navigation.admin.using_version_lock', False)
+    def test_moving_node_version_lock_not_installed_works_without_error(self):
```
Aiky30: """A menu item can be moved without issue if version locking is not installed
"""

tests/test_admin.py
```diff
@@ -131,12 +132,63 @@ def test_get_queryset_filters_by_content_id(self):
...
+        response = self.client.post(self.move_url, data=self.data)
+
+        self.assertEquals(response.status_code, 200)
+
+
 class MenuItemAdminChangeViewTestCase(CMSTestCase, UsefulAssertsMixin):
     def setUp(self):
         self.client.force_login(self.get_superuser())
 
     def test_menuitem_change_view(self):
```
Aiky30: We should fix this and describe what's being tested in this test. 

"""The change view does not show draft versions"""

This test is a perfect example of why the docstring should always be in a test, I had to work out what the test was doing by reading code. That should not be the case, imagine if 10's of tests were failing and you had to work out if the test needed reworking or the code that it checks. If you know "why" something is being tested then you can decide easily which one to change.

tests/test_admin.py
```diff
@@ -592,10 +646,11 @@ def test_menuitem_changelist_redirects_if_not_latest_version(self, mocked_messag
 
 class MenuItemAdminMoveNodeViewTestCase(CMSTestCase):
     def setUp(self):
-        self.client.force_login(self.get_superuser())
+        self.user = self.get_superuser()
+        self.client.force_login(self.user)
 
     def test_menuitem_move_node_smoke_test(self):
```
Aiky30: """A node can be moved"""

tests/test_admin.py
```diff
@@ -616,7 +671,7 @@ def test_menuitem_move_node_smoke_test(self):
 
     @patch("django.contrib.messages.error")
     def test_menuitem_move_node_cant_move_outside_of_root(self, mocked_messages):
```
Aiky30: """A node can't be moved outside of a root node because that would pull it out of a tree altogether
"""

djangocms_navigation/admin.py
```diff
@@ -241,8 +273,15 @@ def move_node(self, request, menu_content_id):
         # Disallow moving of a node on anything other than a draft version
         if self._versioning_enabled:
             menu_content = get_object_or_404(
-                MenuContent._base_manager, id=menu_content_id
+                self.MenuModel._base_manager, id=menu_content_id
             )
+            request.menu_content_id = menu_content_id
+            change_perm = self.has_change_permission(request, menu_content)
+            if not change_perm:
+                msg = 'The item is currently locked or you don\'t have permission to change it'
```
Aiky30: The should use the get text _("")

I think the message should also be different for version locking and when version locking is not installed e.g. 

Version locking enabled: _("The item is currently locked or you don\'t have permission to change it")
Version locking disabled: _("You don\'t have permission to change this item")

djangocms_navigation/admin.py
```diff
@@ -102,32 +116,43 @@ def get_preview_link(self, obj):
 
 
 class MenuItemAdmin(TreeAdmin):
+    MenuModel = MenuContent
     form = MenuItemForm
     change_list_template = "admin/djangocms_navigation/menuitem/change_list.html"
     list_display = ["__str__", "get_object_url"]
 
+    def get_admin_name(self, name, reverse=False):
```
Aiky30: I would say that this should be called get_admin_reverse_url() as it is returning the url and not an admin name which the naming would suggest.

djangocms_navigation/admin.py
```diff
@@ -102,32 +116,43 @@ def get_preview_link(self, obj):
 
 
 class MenuItemAdmin(TreeAdmin):
+    MenuModel = MenuContent
```
Aiky30: this should probably be menu_model, look at form below: form = MenuItemForm

djangocms_navigation/admin.py
```diff
@@ -46,9 +52,22 @@ def url_for_result(self, result):
 
 class MenuContentAdmin(admin.ModelAdmin):
     form = MenuContentForm
+    ItemModel = MenuItem
     list_display = ["title", "get_menuitem_link", "get_preview_link"]
     list_display_links = None
 
+    def get_admin_name(self, name, reverse=False, model=None):
```
Aiky30: Why does this method have the optional model and the same method in MenuItemAdmin does not?

Aiky30: It would be better to create a utility I think, call it something like get_model_admin_reverse in this projects utils.py like: https://github.com/divio/django-cms/blob/develop/cms/utils/urlutils.py#L95

djangocms_navigation/cms_config.py
```diff
@@ -83,6 +86,8 @@ def on_menu_content_archive(version):
 
 
 class NavigationCMSAppConfig(CMSAppConfig):
+    # NOTE what to do in sitemap. Very similar. It will register the correct model
+    # with the config. But navigation models may be different. Settings for this?
```
jonathan-s: @Aiky30 How would you like to solve this, any ideas? The only difference is `navigation_models`.

jonathan-s: I got this.

tests/test_admin.py
jonathan-s: Fixed.

tests/test_admin.py
jonathan-s: Fixed.

djangocms_navigation/utils.py
```diff
@@ -1,15 +1,53 @@
...
+    return url
+
+
 @lru_cache(maxsize=1)
 def supported_models():
     try:
-        app_config = apps.get_app_config("djangocms_navigation")
+        MenuContent = get_model('MENU_MODEL')
+        app_label = MenuContent._meta.app_label
+        app_config = apps.get_app_config(app_label)
```
Aiky30: This is smart, I like this

djangocms_navigation/utils.py
```diff
@@ -19,7 +57,9 @@ def supported_models():
 
 @lru_cache(maxsize=1)
 def supported_content_type_pks():
-    app_config = apps.get_app_config("djangocms_navigation")
+    MenuContent = get_model('MENU_MODEL')
+    app_label = MenuContent._meta.app_label
+    app_config = apps.get_app_config(app_label)
```
Aiky30: I'm thinking that because this is repeated this block should be a utility to get the apps config, an internal `get_app_config_from_model()` Returns app_config

tests/test_utils.py
```diff
@@ -23,23 +23,28 @@ def setUp(self):
...
 
-    def test_supported_models(self):
-        models = ["Foo", "Bar"]
-        app_config = Mock(
-            spec=[], cms_extension=Mock(spec=[], navigation_apps_models=models)
-        )
-        with patch.object(apps, "get_app_config", return_value=app_config):
-            self.assertEqual(supported_models(), models)
+    def test_supported_models_with_installed_apps(self):
+        """Tests that all models registered in cms_config shows
```
Aiky30: Try not to start a test with: "Tests that", we all do it and this comment was said to me: "It's obvious that it's testing that because it's a test"

You can list the business / technical requirement or edge case:
"All models registered in cms_config are shown as supported models"

tests/test_utils.py
```diff
@@ -23,23 +23,28 @@ def setUp(self):
...
+            PollContent: ["text"],
+        }
+        self.assertDictEqual(registered_models, expected_models)
 
     @patch.object(apps, "get_app_config", side_effect=LookupError)
     def test_supported_models_returns_empty_list_on_lookup_error(self, mocked_apps):
         self.assertDictEqual(supported_models(), {})
 
     def test_supported_models_is_cached(self):
-        models = ["Foo", "Bar"]
```
Aiky30: This looks liek something Monika would have done. If it was created by Monika there is a reason that she's mocking here. I won't reject this change but try and understand why she may have mocked here.

djangocms_navigation/constants.py
```diff
@@ -21,8 +21,9 @@ def get_templates():
     return choices
 
 
-PLUGIN_URL_NAME_PREFIX = "djangocms_navigation"
-
-SELECT2_CONTENT_OBJECT_URL_NAME = "{}_select2_content_object".format(
-    PLUGIN_URL_NAME_PREFIX
-)
+def get_select2_url_name():
```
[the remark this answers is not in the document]
jonathan-s: Happy to move this, don't feel strongly about it.

tests/test_utils.py
jonathan-s: Using the app_config as a mock object will get you an error. Generally I'm of the opinion if you can use real objects you should as that will help you avoid bugs. For this particular example I can't find a reason to use a mock for the app_config.  

In this case we're testing the cache and with the approach below it absolutely fulfills that criteria.

djangocms_navigation/admin.py
```diff
@@ -292,10 +313,9 @@ def get_object_url(self, obj):
     @property
     def _versioning_enabled(self):
         """Helper property to check if versioning is enabled for navigation"""
-        return apps.get_app_config(
```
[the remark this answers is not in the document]
jonathan-s: The overview of this patch is that it makes djangocms-navigation re-usable. The current trade-off that's been made is that if you've got several apps that need to use djangocms-navigation AND these apps need to have customized models that will be somewhat difficult. (ie the the global setting of menu model + the above). 

If we need to support djangocms-navigation being re-usable and having multiple apps using customized djangocms-navigation models. I think I would need to spend more time working out the edge cases and write tests that explicitly tests this.  

For reference, this is what the config in sitemap would look like. 
```
class HeadlessSitemapCMSAppConfig(CMSAppConfig):
    navigation_models = {
        # model_class : field(s) to search in menu item form UI
        Page: ["title"],
        Url: ["internal_name"]
    }

    fil_headless_deployment_enabled = True
    deployment_serializer_backends = {MenuContent: SitemapSerializerBackend}
```
If you include `djangocms_versioning_enabled = True` in the headless sitemap you would also need to include. 

```
        versioning = [
            VersionableItem(
                content_model=MenuContent,
                grouper_field_name="menu",
                copy_function=copy_menu_content,
                preview_url=MenuContent.get_preview_url,
                on_publish=on_menu_content_publish,
                on_unpublish=on_menu_content_unpublish,
                on_draft_create=on_menu_content_draft_create,
                on_archive=on_menu_content_archive,
            )
        ]
```

When you do MenuContent would be registered once in the sitemap addon and once in djangocms-navigation which would throw an error if I remember correctly.

djangocms_navigation/utils.py
jonathan-s: Thank you!

jonathan-s:
The following issue cropped up when developing. 
https://github.com/divio/djangocms-url-manager/issues/26

tests/test_views.py
```diff
@@ -1,12 +1,11 @@
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.sites.models import Site
-from django.urls.exceptions import NoReverseMatch
 
 from cms.models import Page, User
 from cms.test_utils.testcases import CMSTestCase
 from cms.utils.urlutils import admin_reverse
 
-from djangocms_navigation.constants import get_select2_url_name
+from djangocms_navigation.utils import get_select2_url_name
```
Aiky30: This should be stacked in alphabetical order with models, then test_utils then utils.

djangocms_navigation/cms_config.py
```diff
@@ -1,67 +1,103 @@
...
+                    grouper_field_name="menu",
+                    copy_function=complete_copy_menu_content,
+                    preview_url=MenuContent.get_preview_url,
+                    on_publish=on_menu_content_publish,
+                    on_unpublish=on_menu_content_unpublish,
+                    on_draft_create=on_menu_content_draft_create,
+                    on_archive=on_menu_content_archive,
+                )
+            ]
+            cms_config.versioning = versioning
```
jonathan-s: Add the following code after line 61

```
            from cms.utils.setup import configure_cms_apps
            versioning_app = apps.get_app_config('djangocms_versioning')
            # will throw an error because configuring versioning apps is not idempotent
            configure_cms_apps([versioning_app])
```

djangocms_navigation/cms_config.py
```diff
@@ -1,71 +1,73 @@
...
 from cms.app_base import CMSAppConfig, CMSAppExtension
 from cms.models import Page
 
-from .utils import purge_menu_cache, get_model
-from .constants import MENU_MODEL, ITEM_MODEL
-
-
-MenuContent = get_model(MENU_MODEL)
-MenuItem = get_model(ITEM_MODEL)
+from .utils import purge_menu_cache
```
Aiky30: tiny change, the .utils import should come after .models

djangocms_navigation/cms_config.py
```diff
@@ -1,71 +1,73 @@
...
+        if isinstance(navigation_app_models, dict):
+            self.navigation_apps_models.update(navigation_app_models)
+        else:
+            raise ImproperlyConfigured(
+                "navigation configuration must be a dictionary object"
+            )
+
 
-def copy_menu_content(original_content):
+def copy_menu_content(menu_model, item_model, original_content):
```
Aiky30: We don't know if this could have been reused by other apps, because of this I think we should make the additional attributes optional: `def copy_menu_content(original_content, menu_model=MenuContent, item_model=MenuItem):`. Partial won't work in this scenario and we would have to wrap the function. Any ideas of how we could do that?

The solution designed here will be used as an example for any addons that reuse copy methods. I think it's important that we make them backwards compatible.

jonathan-s: You are right that it's better to make this backwards compatible. Making the signature the above looks to be the correct choice. 

However I think partial still does work. Take a look at the python docs: https://docs.python.org/3.6/library/functools.html#functools.partial

It should work with keyword arguments as well.
```
>>> from functools import partial
>>> basetwo = partial(int, base=2)
>>> basetwo.__doc__ = 'Convert base 2 string to an int.'
>>> basetwo('10010')
```

jonathan-s: You are right that it's better to make this backwards compatible. Making the signature the above looks to be the correct choice. 

However I think partial still does work. Take a look at the python docs: https://docs.python.org/3.6/library/functools.html#functools.partial

It should work with keyword arguments as well.
```
>>> from functools import partial
>>> basetwo = partial(int, base=2)
>>> basetwo.__doc__ = 'Convert base 2 string to an int.'
>>> basetwo('10010')
```

jonathan-s: Looks like I was wrong. How about we solve it this way?

```
def _copy_menu_content(original_content, menu_model, item_model):
   # snip 
    return new_content


complete_copy_menu_content = partial(
    _copy_menu_content,
    MenuContent,
    MenuItem
)


def copy_menu_content(original_content):
    """Copy the MenuContent object and deepcopy its menu items."""
    return complete_copy_menu_content(original_content)
```

jonathan-s: That way `copy_menu_content` keeps its original signature.

djangocms_navigation/utils.py
```diff
@@ -2,9 +2,34 @@
...
+# FIXME: Reuse the django cms reverse util: admin_reverse()
+def reverse_admin_name(model, name, args=None, kwargs=None):
+    name = get_admin_name(model, name)
+    url = reverse('admin:{}'.format(name), args=args, kwargs=kwargs)
+    return url
+
+
+def get_select2_url_name():
+    url_name = "{}_select2_content_object".format(
+        MenuContent._meta.app_label
```
Aiky30: Make the model configurable for this method.

PR #53, opened 2019-11-06
Configurable models for reuse

Aiky30:
@jonathan-s Had created additional work that was deemed too many changes to take into a production release. The code has been reworked to make the code reusable, but in a way that it takes over the code and two cannot co exist.


tests/test_cms_config.py
```diff
@@ -65,22 +63,6 @@ def test_valid_cms_config_parameter(self):
             self.assertTrue(TestModel4 in register_model)
 
 
-class NavigationIntegrationTestCase(TestCase):
```
Aiky30: Shouldn't be removed!!!

jonathan-s:
@Aiky30 I believe this PR can be closed as it has been implemented in #60

PR #56, opened 2020-01-22
 Version locked items can't be moved or changed - cherrypicked

jonathan-s:
Cherrypicked changes explicitly to fix this issue from #51

### DIFF
--- a/djangocms_navigation/admin.py
+++ b/djangocms_navigation/admin.py
@@ -29,6 +29,17 @@ try:
 except ImportError:
     pass
 
+try:
+    from djangocms_version_locking.helpers import content_is_unlocked_for_user
+    using_version_lock = True
+    LOCK_MESSAGE = _(
+        "The item is currently locked or you don't "
+        "have permission to change it"
+    )
+except ImportError:
+    using_version_lock = False
+    LOCK_MESSAGE = _("You don't have permission to change this item")
+
 
 class MenuItemChangeList(ChangeList):
     def __init__(self, request, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


djangocms_navigation/admin.py
```diff
@@ -165,23 +176,30 @@ def change_view(
...
-                kwargs={"menu_content_id": menu_content_id},
+
+        if self._versioning_enabled:
+            menu_content = get_object_or_404(
+                self.MenuModel._base_manager, id=menu_content_id
             )
 
+            change_perm = self.has_change_permission(request, menu_content)
+            if not change_perm:
+                messages.error(request, 'You don\'t have permission to edit or it is locked')
```
Aiky30: Is there a reason that this message is not using: LOCK_MESSAGE?

tests/test_admin.py
```diff
@@ -171,12 +172,66 @@ def test_get_queryset_filters_by_content_id(self):
         )
 
 
+class MenuItemAdminVersionLocked(CMSTestCase, UsefulAssertsMixin):
+
+    def setUp(self):
+        # who is author..
```
Aiky30: What is the following comment for? # who is author

jonathan-s: I think it was for the version of the factory. But taking a look at how another version factory looked like it looks the same. So I deleted this.

jonathan-s:
@Aiky30 Sorry it took so long to get back to this. I've fixed the issues here.

djangocms_navigation/admin.py
```diff
@@ -165,23 +176,30 @@ def change_view(
...
-                    return HttpResponseRedirect(version_list_url(menu_content))
-                # purge menu cache
-                purge_menu_cache(site_id=menu_content.menu.site_id)
-            extra_context["list_url"] = reverse(
-                "admin:djangocms_navigation_menuitem_list",
-                kwargs={"menu_content_id": menu_content_id},
+
+        if self._versioning_enabled:
+            menu_content = get_object_or_404(
+                self.MenuModel._base_manager, id=menu_content_id
```
Aiky30: I can't see where self.MenuModel is defined. Is this defined anywhere?

Aiky30: @jonathan-s I couldn't see where this was used, it's possibly left over from changes in the other PR for making the models reusable: self.MenuModel

PR #57, opened 2020-02-14
Fixed some minor grammatical errors 

adam-murray:
The readme was confusing because of the grammatical issues, I have fixed them so it's easier to understand.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,11 +40,6 @@ You can run all the tests by executing:
     pip install -r tests/requirements.txt
     python tests.settings.py
 
-    # Or on MacOS use Pyenv:
-    pyenv virtualenv myvirtualenvironmentname
-    pyenv activate myvirtualenvironmentname
-
-
 App Integration
 ===============
 
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,7 @@ Running Tests
 
 You can run all the tests by executing:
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -32,24 +32,28 @@ Running Tests
...
     source venv/bin/activate
     python setup.py test
 
     # Alternatively you could run the test runner stand-alone
     pip install -r tests/requirements.txt
     python tests.settings.py
 
+    # Or on MacOS use Pyenv:
+    pyenv virtualenv myvirtualenvironmentname
+    pyenv activate myvirtualenvironmentname
```
jonathan-s: When using python3 pyenv is redundant.

Aiky30: Agreed, the above '-m venv' is built into python which makes it platform independent and works for Mac, Linux and Windows

README.rst
```diff
@@ -32,24 +32,28 @@ Running Tests
 
 You can run all the tests by executing:
 
-    python -m venv venv
+    python -m venv myvirtualenvironmentname
```
Aiky30: venv is fine here, most Pythyon tutorials use 'venv' so that's consistent with other packages and Python tutorials.

README.rst
Aiky30: Please remove

PR #60, opened 2020-02-28
Make the packages codebase reusable for other packages to compliment or fully override

Aiky30:
Part of a joint effort between @jonathan-s and @Aiky30 to make the application code reusable in other packages to allow the other packages to create opinionated navigation with the least amount of repetitive code.

Code and ideas taken from:
https://github.com/FidelityInternational/djangocms-navigation/pull/51
https://github.com/FidelityInternational/djangocms-navigation/pull/53


djangocms_navigation/admin.py
```diff
@@ -329,10 +345,11 @@ def get_object_url(self, obj):
...
+
         return apps.get_app_config(
-            "djangocms_navigation"
+            self.model._meta.app_label
         ).cms_config.djangocms_versioning_enabled
 
 
-admin.site.register(MenuItem, MenuItemAdmin)
 admin.site.register(MenuContent, MenuContentAdmin)
+admin.site.register(MenuItem, MenuItemAdmin)
```
jonathan-s: Perhaps it would make sense to have a config with a sane default for this? Ie if you only are re-using the MenuItemAdmin and MenuContentAdmin you wouldn't really want to register MenuItem and MenuContent.

Aiky30: This package can be used as is, it just so happens that there are requirements that are very specific to FIL that need to be catered for in different applications. It's still valid to have the models registered.

PR #63, opened 2020-03-19
Support django 2.2

jonathan-s:
Depends on:
- https://github.com/divio/djangocms-versioning/pull/199
- https://github.com/divio/djangocms-moderation/pull/168

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,9 +4,9 @@ import djangocms_navigation
 
 
 INSTALL_REQUIREMENTS = [
-    "Django>=1.11,<2.2",
-    "django-cms>=3.5.0",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -4,8 +4,7 @@
 
 
 INSTALL_REQUIREMENTS = [
-    "Django>=1.11,<2.2",
-    "django-cms>=3.5.0",
```
Aiky30: @jonathan-s is djangocms not required for testing locally?

jonathan-s: Apologies, deleted the entire line by mistake. If you check the most recent diff it has been added in.

PR #68, opened 2020-04-06
Added language migration for existing databases with no language value

adam-murray:


djangocms_navigation/migrations/0007_language_migration.py
```diff
@@ -0,0 +1,36 @@
...
+from cms.utils.i18n import get_current_language
+
+logger = logging.getLogger(__name__)
+
+
+def add_navigation_languages(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    navigation_queryset = apps.get_model("djangocms_navigation", "MenuContent").using(db_alias).filter(language_exact="")
+    for model in navigation_queryset:
+        model.language = get_current_language()
```
Aiky30, line 17: This needs to be done on a per site bases, there is a method that helps to get the language for the site, I saw it yesterday, I'll comment below when I find it again.

djangocms_navigation/models.py
```diff
@@ -38,6 +39,9 @@ def root_id(self):
 
 
 class AbstractMenuContent(models.Model):
+    # There is an additional migration (0007_language_migration) to handle existing unpopulated database entries
+    language = models.CharField(_("language"), max_length=15, db_index=True, default=get_default_language)
```
Aiky30, line 43: This could possibly need to use the get default language for a site too.

djangocms_navigation/filters.py
```diff
@@ -0,0 +1,33 @@
+from django.contrib import admin
+from django.utils.translation import ugettext_lazy as _
+
+from cms.utils.i18n import get_language_tuple, get_site_language_from_request
+
+
+class LanguageFilter(admin.SimpleListFilter):
+    title = _("language")
```
Aiky30, line 8: The language filter will need to be registered with the admin, an example of where this has been done before: https://github.com/FidelityInternational/djangocms-pageadmin/blob/master/djangocms_pageadmin/admin.py#L48

djangocms_navigation/migrations/0008_language_migration.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db import migrations
+
+from cms.utils.i18n import get_default_language_for_site
+
+logger = logging.getLogger(__name__)
+
+
+def add_navigation_languages(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    navigation_queryset = apps.get_model("djangocms_navigation", "MenuContent").using(db_alias).filter(language_exact="")
```
Aiky30, line 15: This is erroring: 

```
  File "/virtualenv/lib/python3.6/site-packages/djangocms_navigation/migrations/0008_language_migration.py", line 15, in add_navigation_languages
    navigation_queryset = apps.get_model("djangocms_navigation", "MenuContent").using(db_alias).filter(language_exact="")
AttributeError: type object 'MenuContent' has no attribute 'using'
```

Aiky30: Try and be a little cleaner and explicit with this, what I mean here is to make it easier to digest what is happening rather than trying to do it all on one line:

```
MenuContent = apps.get_model("djangocms_navigation", "MenuContent")
navigation_queryset = MenuContent.objects.using(db_alias).filter(language_exact="")
```

Notice the `.objects.using(` it was the missing objects that threw the error.

djangocms_navigation/migrations/0008_language_migration.py
```diff
@@ -0,0 +1,36 @@
...
+
+from cms.utils.i18n import get_default_language_for_site
+
+logger = logging.getLogger(__name__)
+
+
+def add_navigation_languages(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    MenuContent = apps.get_model("djangocms_navigation", "MenuContent")
+    navigation_queryset = MenuContent.objects.using(db_alias).filter(language_exact="")
```
Aiky30, line 16: New error:
```
    "Choices are: %s" % (name, ", ".join(available)))
django.core.exceptions.FieldError: Cannot resolve keyword 'language_exact' into field. Choices are: id, menu, menu_id, root, root_id
```

language_exact should be: language__exact

Please get this running the migration successfully using the gaps test project before I look again.

djangocms_navigation/forms.py
```diff
@@ -20,10 +21,11 @@ class Meta:
 
 class MenuContentForm(forms.ModelForm):
     title = forms.CharField(label=_("Menu Title"), max_length=100)
+    language = forms.ChoiceField(choices=tuple(get_language_tuple()))
```
Aiky30, line 24: Last change, we're going to need to filter the list of languages to only those that are available for the site: https://github.com/divio/django-cms/blob/develop/cms/utils/i18n.py#L94

Aiky30: What could happen is that a language could be added that could never be used on a site.

djangocms_navigation/migrations/0008_menucontent_language.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('djangocms_navigation', '0007_auto_20200302_0853'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='menucontent',
+            name='language',
+            field=models.CharField(db_index=True, default=cms.utils.i18n.get_default_language, max_length=15, verbose_name='language'),
```
Aiky30, line 17: I don't know how I didn't see this before. The next data migration will never run because we always set a default language here. I think this should be default="". It would have been nice to use default=get_default_language_for_site but that is not possible because we can't access model.site unfortunately. 

https://eligaservices.atlassian.net/wiki/spaces/DEV/pages/54526331/Glossary+of+Terms+Terminology

PR #72, opened 2020-07-16
Add helper wrapper to search content object from node tree

NarenderRajuB:
Created a Helper to find a content_object in a Navigation tree for use in various scenarios and changes coming through.

### DIFF
--- a/djangocms_navigation/helpers.py
+++ b/djangocms_navigation/helpers.py
@@ -1,0 +1,21 @@
+from django.contrib.contenttypes.models import ContentType
+
+from .models import MenuItem
+
+
+def search_content_object_from_node_tree(menu_content, content_object):
+    """
+    Search for the content_object in menu_content node tree
+    Input:  Menu content node tree , Content object to be searched
+    Output: Node mapped to Content object if found in Node Tree else return False
+    """
+
+    content_object_content_type = ContentType.objects.get_for_model(content_object)
+    root = MenuItem.get_root_nodes().filter(menucontent=menu_content).first()
+    if root:
+        child_nodes = root.get_descendants()
+        search_node = child_nodes.filter(content_type=content_object_content_type, object_id=content_object.pk).first()
+        if search_node:
+            return search_node
+
+    return False
--- a/djangocms_navigation/helpers.py
+++ b/djangocms_navigation/helpers.py
@@ -3,16 +3,16 @@ from django.contrib.contenttypes.models import ContentType
 from .models import MenuItem
 
 
-def search_content_object_from_node_tree(menu_content, content_object):
+def get_content_object_from_navigation_tree(menu_content, content_object):
     """
     Search for the content_object in menu_content node tree
     Input:  Menu content node tree , Content object to be searched
     Output: Node mapped to Content object if found in Node Tree else return False
     """
 
-    content_object_content_type = ContentType.objects.get_for_model(content_object)
     root = MenuItem.get_root_nodes().filter(menucontent=menu_content).first()
     if root:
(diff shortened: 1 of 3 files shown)


djangocms_navigation/helpers.py
```diff
@@ -0,0 +1,21 @@
+from django.contrib.contenttypes.models import ContentType
+
+from .models import MenuItem
+
+
+def get_content_object_from_navigation_tree(menu_content, content_object):
```
Aiky30, line 6: Sorry mate, I hate to be a pain. The name of the helper is very important as it will be used a lot in the future and documented. 

1. A better name that describes what it is doing is:

`get_navigation_node_for_content_object`

2. There needs to be an additional argument with a default as the package is reused in FIL: node_model=MenuItem

3. The docstring should describe the in and out the same as the other packages, versioning is the best example of this: https://github.com/divio/djangocms-versioning/blob/master/djangocms_versioning/helpers.py#L21:

    """
    Find a navigation node that contains a content_object in a Navigation menu

    :param menu_content: A MenuContent instance
    :param content_object:  A content object registered in the cms_config
    :param node_model:  A model used for a navigation item
    :return: A navigation node or False
    """

tests/test_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class NavigationContentTypeSearchTestCase(CMSTestCase):
+
+    def setUp(self):
+        self.language = settings.LANGUAGES[0][0]
+
+    def test_page_content_type_in_node_tree(self):
+        """
+        Test to ensure the correct node mapped to poll content is returned by the helper"
```
Aiky30, line 17: Reword to: "The correct node mapped to poll content is returned by the helper"

tests/test_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+        grandchild1 = factories.ChildMenuItemFactory(parent=grandchild)
+        grandchild2 = factories.ChildMenuItemFactory(parent=grandchild1, content=page_content)
+
+        result = get_navigation_node_for_content_object(menu_contents, page_content)
+
+        self.assertEqual(result, grandchild2)
+
+    def test_content_object_not_found_in_node_tree(self):
+        """
+        Test to ensure False if returned  by the helper when no node is mapped to content in node tree
```
Aiky30, line 50: False is returned by the helper when no node is mapped to content in the node tree

tests/test_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+        grandchild1 = factories.ChildMenuItemFactory(parent=grandchild)
+        factories.ChildMenuItemFactory(parent=grandchild1)
+
+        result = get_navigation_node_for_content_object(menu_contents, page_content)
+
+        self.assertFalse(result)
+
+    def test_content_object_mapped_to_more_than_node(self):
+        """
+        Test to ensure first node found returned by helper when more than one node is mapped to content object
```
Aiky30, line 73: The first node found returned is returned by the helper when more than one node is mapped to a content object in a node tree

tests/test_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+        grandchild1 = factories.ChildMenuItemFactory(parent=grandchild)
+        factories.ChildMenuItemFactory(parent=grandchild1, content=page_content)
+
+        result = get_navigation_node_for_content_object(menu_contents, page_content)
+
+        self.assertEqual(result, grandchild)
+
+    def test_to_search_pollcontent_in_node_tree(self):
+        """
+        Test to ensure correct node mapped to poll content is returned from node tree search
```
Aiky30, line 92: The correct node mapped to poll content is returned from node tree search

PR #74, opened 2020-07-22
OPENCMS-921 Implement soft_root functionality in djangocms_navigation

NarenderRajuB:
Implement soft_root functionality in djangocms_navigation


tests/test_cms_menus.py
```diff
@@ -190,3 +246,187 @@ def test_get_roots_with_versioning_disabled(self):
...
+                  6:bbb
+        """
+        self.root.soft_root = True
+        page_url = self.aaa_pagecontent.page.get_absolute_url()
+        context = self.get_context(page_url, page=self.root_pagecontent.page)
+        tpl = Template("{% load menu_tags %}{% show_menu 0 100 100 100 %}")
+        tpl.render(context)
+        soft_root = context['children']
+        mock_tree = [
+            AttributeObject(title=self.root.title, level=0, children=[
```
Aiky30, line 420: This test does not work for softroots, it's exactly the same as the test above it. I don't think that the original test in the cms is anygood either. Why would you test for a softroot on the root, it's already the root!

Try this, Make self.aaa the soft_root, and then use the page_url as 
```
page = self.aaa1_pagecontent.page
context = self.get_context(page.get_absolute_url(), page=page)
```

djangocms_navigation/cms_menus.py
```diff
@@ -11,6 +11,20 @@
...
 
+class MenuItemNavigationNode(NavigationNode):
+
+    def __init__(self, *args, **kwargs):
+        self.content = kwargs.pop('content')
+        super().__init__(*args, **kwargs)
+
+    def is_selected(self, request):
+        try:
+            content = request.current_page
```
Aiky30, line 22: current_page in the real world is page. It is now returning false. No test is actually testing the softroot functionality, this prove the fact.

$ type(request.current_page)
Page

$ type(self.content)
PageContent

tests/test_helpers.py
```diff
@@ -1,11 +1,17 @@
...
 
 from cms.test_utils.testcases import CMSTestCase
+from cms.test_utils.util.fuzzy_int import FuzzyInt
 
 from djangocms_navigation.helpers import get_navigation_node_for_content_object
 from djangocms_navigation.test_utils import factories
 from djangocms_navigation.test_utils.polls.models import Poll, PollContent
 
+try:
+    from djangocms_versioning.constants import PUBLISHED
```
Aiky30, line 11: Versioning is very much an integral part of Navigation, there's no need for this try catch.

tests/test_cms_menus.py
Aiky30: Setting self.navnode.soft_root = True is always seen as False in is_selected. You will need to copy all of the nodes from the setup and create them here.

NarenderRajuB: Updated the tests with correct page request and expected results.

djangocms_navigation/cms_menus.py
NarenderRajuB: Updated as suggested

tests/test_helpers.py
NarenderRajuB: Done updated

tests/test_cms_menus.py
```diff
@@ -190,3 +247,446 @@ def test_get_roots_with_versioning_disabled(self):
...
+        """
+        msg = '%r != %r with %r, %r' % (len(a), len(b), a, b)
+        self.assertEqual(len(a), len(b), msg)
+        for n1, n2 in zip(a, b):
+            for attr in attrs:
+                a1 = getattr(n1, attr)
+                a2 = getattr(n2, attr)
+                msg = '%r != %r with %r, %r (%s)' % (a1, a2, n1, n2, attr)
+                self.assertEqual(a1, a2, msg)
+            self.assertTreeQuality(n1.children, n2.children, 'level', 'title')
```
Aiky30, line 296: This should be: self.assertTreeQuality(n1.children, n2.children, *attrs), that will send the attrs back in.

PR #88, opened 2020-09-18
OPENCMS-1039 Add edit_mode_active flag to render draft only menu nodes

NarenderRajuB:
fix menu rendering for menu item nodes for draft/published mode

### DIFF
--- a/tests/test_cms_menus.py
+++ b/tests/test_cms_menus.py
@@ -16,6 +16,7 @@ from djangocms_versioning.constants import (
 )
 
 from djangocms_navigation.cms_menus import CMSMenu
+from djangocms_navigation.models import MenuItem
 from djangocms_navigation.test_utils import factories
 
 from .utils import disable_versioning_for_navigation
@@ -252,7 +253,10 @@ class CMSMenuTestCase(CMSTestCase):
             list(roots), [menucontent_1.root, menucontent_2.root, menucontent_3.root]
         )
 
-    def test_menu_nodes_in_page_view_draft_mode(self):
+    def test_draft_menu_on_draft_page(self):
+        """
+        Draft menu changes on draft page rendering.
+        """
         menu_cont_draft = factories.MenuContentWithVersionFactory(version__state=DRAFT, language=self.language)
         menu_cont_published = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
         pagecontent_aaa = factories.PageContentWithVersionFactory(
@@ -284,7 +288,10 @@ class CMSMenuTestCase(CMSTestCase):
         self.assertIn(child.title, str(response.content))
         self.assertNotIn(child_1.title, str(response.content))
 
-    def test_menu_nodes_in_page_view_publish_mode(self):
+    def test_draft_menu_on_published_page(self):
+        """
+         Draft menu changes on published page
+        """
         menu_cont_draft = factories.MenuContentWithVersionFactory(version__state=DRAFT, language=self.language)
         menu_cont_published = factories.MenuContentWithVersionFactory(version__state=PUBLISHED,
                                                                       language=self.language)
@@ -316,7 +323,10 @@ class CMSMenuTestCase(CMSTestCase):
         self.assertIn(child_1.title, str(response.content))
         self.assertNotIn(child.title, str(response.content))
 
-    def test_menu_nodes_in_page_view_preview_mode(self):
+    def test_draft_menu_in_draft_preview_mode(self):
+        """
+        Draft menu changes on draft poge in preview mode
+        """
         menu_cont_draft = factories.MenuContentWithVersionFactory(version__state=DRAFT, language=self.language)
         menu_cont_published = factories.MenuContentWithVersionFactory(version__state=PUBLISHED,
                                                                       language=self.language)
@@ -380,6 +390,7 @@ class SoftrootTests(CMSTestCase):
     def setUp(self):
         self.language = 'en'
         self.client.force_login(self.get_superuser())
+        self.user = factories.UserFactory()
         self.root_pagecontent = factories.PageContentWithVersionFactory(
             language=self.language,
             version__created_by=self.get_superuser(),
@@ -824,6 +835,38 @@ class SoftrootTests(CMSTestCase):
         ]
 
         self.assertTreeQuality(soft_root, mock_tree, 'title', 'level')
+
+    def test_draft_menu_item_atributechanges_on_published_page_menu(self):
+        """
+        Draft menu item attribute changes render changes in draft menucontent version only.
+        """
+
+        original_version = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
+        original_root = factories.ChildMenuItemFactory(
+            parent=original_version.root, content=self.root_pagecontent
+        )
+
+        original_child = factories.ChildMenuItemFactory(parent=original_root, content=self.aaa_pagecontent.page)
+        original_grandchild = factories.ChildMenuItemFactory(parent=original_child, content=self.bbb_pagecontent.page)
+
+        published_version = original_version.versions.get()
+        new_version = published_version.copy(self.user)
+
+        new_root = new_version.content.root
+        new_child = MenuItem.objects.get(path=new_root.path + original_child.path[4:])
+        new_child.soft_root = True
+        new_child.hidden = True
+
+        page = self.bbb_pagecontent.page
+        context = self.get_context(page.get_absolute_url(), page=page)
+        tpl = Template("{% load menu_tags %}{% show_menu 0 100 100 100 %}")
+        tpl.render(context)
+        published_root = context['children']
+
+        self.assertFalse(published_root[0].children[0].attr.get("soft_root"))
+        self.assertEqual(original_child.soft_root, published_root[0].children[0].attr.get("soft_root"))
+        self.assertTrue(published_root[0].children[0].visible)
+
 
     def test_basic_projects_softroot_rendering_nodes(self):
         """
--- a/tests/test_cms_menus.py
+++ b/tests/test_cms_menus.py
@@ -257,27 +257,38 @@ class CMSMenuTestCase(CMSTestCase):
         """
         Ensure that a draft page renders a draft menu when it exists.
         """
-        menu_cont_draft = factories.MenuContentWithVersionFactory(version__state=DRAFT, language=self.language)
-        menu_cont_published = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
+        menu = factories.MenuFactory()
+        menu_cont_published = factories.MenuContentWithVersionFactory(
+            menu=menu,
+            version__state=PUBLISHED,
+            language=self.language
+        )
+        menu_cont_draft = factories.MenuContentWithVersionFactory(
+            menu=menu,
+            version__state=DRAFT,
+            language=self.language
+        )
+        page = factories.PageFactory()
         pagecontent_published = factories.PageContentWithVersionFactory(
+            page=page,
             language=self.language,
             version__created_by=self.get_superuser(),
-            title="test",
-            menu_title="test",
-            page_title="test",
+            version__state=PUBLISHED,
+        )
+        pagecontent_draft = factories.PageContentWithVersionFactory(
+            page=page,
+            language=self.language,
+            version__created_by=self.get_superuser(),
             version__state=DRAFT,
         )
-        version = pagecontent_published.versions.get()
-        version.publish(self.get_superuser())
-        published_version_pagecontent = pagecontent_published.versions.get()
-
-        pagecontent_draft_version = published_version_pagecontent.copy(self.get_superuser())
-        pagecontent_draft = pagecontent_draft_version.content
         draft_child = factories.ChildMenuItemFactory(parent=menu_cont_draft.root, content=pagecontent_draft.page)
         published_child = factories.ChildMenuItemFactory(
             parent=menu_cont_published.root,
             content=pagecontent_published.page
         )
+
+        self.assertEqual(pagecontent_draft.page, pagecontent_published.page)
+        self.assertEqual(menu_cont_draft.menu, menu_cont_published.menu)
 
         # Node added in draft menu version is rendered in page draft view only
         draft_page_endpoint = get_object_edit_url(pagecontent_draft)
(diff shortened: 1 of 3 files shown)


tests/test_plugin.py
```diff
@@ -412,8 +421,8 @@ def test_menu_cache_invalidate_after_menucontent_archive(self):
             language=self.language, version__created_by=self.get_superuser()
         )
         placeholder = factories.PlaceholderFactory(source=page_content)
-        menu_content_version = factories.MenuVersionFactory(content__language=self.language)
-        menu_content = menu_content_version.content
+        menu_content_version = factories.MenuVersionFactory.create_batch(2, content__language=self.language)
```
Aiky30, line 424: I think this shouldn't create a batch of 2, it's possibly broken the tests because they are testing the one version and cache!

tests/test_plugin.py
```diff
@@ -321,8 +329,8 @@ def test_menu_cache_invalidate_after_menucontent_publish(self):
...
```
Aiky30, line 332: I think this shouldn't create a batch of 2, it's possibly broken the tests because they are testing the one version and cache!

tests/test_plugin.py
```diff
@@ -18,6 +19,9 @@ class NavigationSelectorTestCase(TestCase):
     def setUp(self):
         self.selector = NavigationSelector(None)
         self.request = RequestFactory().get("/")
+        self.user = factories.UserFactory()
```
Aiky30, line 22: self.user isn't used anywhere, it could just be:

`self.request.user = factories.UserFactory()`

Aiky30 (changes requested):

tests/test_plugin.py
NarenderRajuB: updated tests

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
             list(roots), [menucontent_1.root, menucontent_2.root, menucontent_3.root]
         )
 
+    def test_draft_menu_on_draft_page(self):
+        """
+        Draft menu changes on draft page rendering.
```
Aiky30, line 258:
```suggestion
        Ensure that a draft page renders a draft menu when it exists.
```

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+        pagecontent_bbb = factories.PageContentWithVersionFactory(
+            language=self.language,
+            version__created_by=self.get_superuser(),
+            title="bbb",
+            menu_title="bbb",
+            page_title="bbb",
+            version__state=PUBLISHED,
+        )
+
+        child = factories.ChildMenuItemFactory(parent=menu_cont_draft.root, content=pagecontent_aaa.page)
```
Aiky30, line 315:
```suggestion
        draft_child = factories.ChildMenuItemFactory(parent=menu_cont_draft.root, content=pagecontent_aaa.page)
```

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+        child_1 = factories.ChildMenuItemFactory(parent=menu_cont_published.root, content=pagecontent_bbb.page)
```
Aiky30, line 316:
```suggestion
        published_child = factories.ChildMenuItemFactory(parent=menu_cont_published.root, content=pagecontent_bbb.page)
```

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+
+        # Node added in draft menu version is not rendered in published view, only published menu nodes are rendered
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.get(pagecontent_bbb.page.get_absolute_url())
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(child_1.title, str(response.content))
```
Aiky30, line 323: This is now easier to follow as: 

```suggestion
        self.assertIn(published_child.title, str(response.content))
```

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+        self.assertNotIn(child.title, str(response.content))
```
Aiky30, line 324:
```suggestion
        self.assertNotIn(draft_child.title, str(response.content))
```

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.get(draft_page_endpoint)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(child.title, str(response.content))
+        self.assertNotIn(child_1.title, str(response.content))
+
+    def test_draft_menu_on_published_page(self):
+        """
+         Draft menu changes on published page
```
Aiky30, line 293: The comment is basically the same as the test name, add value or don't add it at all.

```suggestion
         Ensure that a published page only shows a published menu
```

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+        """
+        menu_cont_draft = factories.MenuContentWithVersionFactory(version__state=DRAFT, language=self.language)
+        menu_cont_published = factories.MenuContentWithVersionFactory(version__state=PUBLISHED,
+                                                                      language=self.language)
+        pagecontent_aaa = factories.PageContentWithVersionFactory(
```
Aiky30, line 298:
```suggestion
        pagecontent_draft = factories.PageContentWithVersionFactory(
```

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+            language=self.language,
+            version__created_by=self.get_superuser(),
+            title="aaa",
+            menu_title="aaa",
+            page_title="aaa",
+            version__state=DRAFT,
+        )
+        pagecontent_bbb = factories.PageContentWithVersionFactory(
```
Aiky30, line 306:
```suggestion
        pagecontent_published = factories.PageContentWithVersionFactory(
```

tests/test_cms_menus.py
Aiky30, line 315:
```suggestion
        child = factories.ChildMenuItemFactory(parent=menu_cont_draft.root, content=pagecontent_draft.page)
```

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
             list(roots), [menucontent_1.root, menucontent_2.root, menucontent_3.root]
         )
 
+    def test_draft_menu_on_draft_page(self):
+        """
+        Draft menu changes on draft page rendering.
+        """
+        menu_cont_draft = factories.MenuContentWithVersionFactory(version__state=DRAFT, language=self.language)
+        menu_cont_published = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
+        pagecontent_aaa = factories.PageContentWithVersionFactory(
```
Aiky30, line 262: This test will only work if the pages share the same grouper (page). The published version has to be added first too as this is logically how the versions are created.

tests/test_cms_menus.py
Aiky30, line 298: This test will only work if the pages share the same grouper (page). The published version has to be added first too as this is logically how the versions are created.

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(child_1.title, str(response.content))
+        self.assertNotIn(child.title, str(response.content))
+
+    def test_draft_menu_in_preview_mode(self):
+        """
+        rendering draft poge in preview mode will show published menu
+        """
+        menu_cont_published = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
+        pagecontent_aaa = factories.PageContentWithVersionFactory(
```
Aiky30, line 331: This test will only work if the pages share the same grouper (page). The published version has to be added first too as this is logically how the versions are created.

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+            language=self.language,
+            version__created_by=self.get_superuser(),
+            title="aaa",
+            menu_title="aaa",
+            page_title="aaa",
+            version__state=DRAFT,
+        )
+        pagecontent_bbb = factories.PageContentWithVersionFactory(
```
Aiky30, line 270: pagecontent_bbb made sense for the menu where they where logical but here it doesn't really make sense. Make it easy for the reader to follow things, pagecontent_published / draft helps here as you aren't changing states.

tests/test_cms_menus.py
```diff
@@ -250,6 +253,115 @@ def test_get_roots_with_versioning_disabled(self):
...
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.get(pagecontent_bbb.page.get_absolute_url())
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(child_1.title, str(response.content))
+        self.assertNotIn(child.title, str(response.content))
+
+    def test_draft_menu_in_preview_mode(self):
+        """
+        rendering draft poge in preview mode will show published menu
```
Aiky30, line 328: This is better than the ones above.

tests/test_cms_menus.py
```diff
@@ -726,6 +839,37 @@ def test_menu_with_softroot_rendering_nested_softroot_child(self):
 
         self.assertTreeQuality(soft_root, mock_tree, 'title', 'level')
 
+    def test_draft_menu_item_atributechanges_on_published_page_menu(self):
+        """
+        Draft menu item attribute changes render changes in draft menucontent version only.
```
Aiky30, line 844: duplicate "changes" in the comment

```suggestion
        A published page only shows published menu when a draft version exists with changes to the attributes.
```

tests/test_cms_menus.py
```diff
@@ -726,6 +839,37 @@ def test_menu_with_softroot_rendering_nested_softroot_child(self):
...
+        """
+
+        original_version = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
+        original_root = factories.ChildMenuItemFactory(
+            parent=original_version.root, content=self.root_pagecontent
+        )
+
+        original_child = factories.ChildMenuItemFactory(parent=original_root, content=self.aaa_pagecontent.page)
```
Aiky30, line 852: Because we are testing attribute changes we need to explicitly set them on each version, this version should set them to False if we are testing them as True further down.

tests/test_plugin.py
```diff
@@ -187,6 +191,8 @@ def test_crud_for_navigation_plugin_when_versioning_enabled(self):
         # so the page can be viewed
         version = page_content.versions.get()
         version.publish(self.get_superuser())
+        menu_version = menu_content2.versions.get()
+        menu_version.publish(self.get_superuser())
         # TODO: Possibly there's a bug where MenuItem objects from
```
Aiky30, line 196: So this TODO message is now redundant as this is now fixed.

tests/test_plugin.py
```diff
@@ -333,18 +341,14 @@ def test_menu_cache_invalidate_after_menucontent_publish(self):
...
-        # Now publish the page content containing the plugin,
-        # so the page can be viewed
-        version = page_content.versions.get()
-        version.publish(self.get_superuser())
-
         # Making sure there is no cachekey existed before rendering page
         cache_key = CacheKey.objects.all().count()
         self.assertEqual(cache_key, 0)
 
-        # And view the page
```
Aiky30, line 345: I'm sure that this test should work because the version was published and the endpoint was a published endpoint!

tests/test_plugin.py
```diff
@@ -424,17 +428,13 @@ def test_menu_cache_invalidate_after_menucontent_archive(self):
...
 
-        # Now publish the page content containing the plugin,
-        # so the page can be viewed
-        version = page_content.versions.get()
-        version.publish(self.get_superuser())
-
         # Making sure there is no cachekey existed before rendering page
         cache_key = CacheKey.objects.all().count()
         self.assertEqual(cache_key, 0)
-        # And view the page
```
Aiky30, line 435: I'm sure that this test should work because the version was published and the endpoint was a published endpoint!

Aiky30 (changes requested):
A few things to address regarding comments, variable names and some test logic that may need to be put back.

tests/test_plugin.py
NarenderRajuB: # https://github.com/divio/django-cms/blob/2daeb7d63cb5fee49575a834d0f23669ce46144e/cms/tests/test_plugins.py#L160 here we  were trying to add a plugin to a draft page.  test above renders  a draft page after adding the plugin which increases the cachecount. earlier before my changes were added.

tests/test_cms_menus.py
NarenderRajuB: Updated

tests/test_cms_menus.py
```diff
@@ -250,6 +253,124 @@ def test_get_roots_with_versioning_disabled(self):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(published_child.title, str(response.content))
+        self.assertNotIn(draft_child.title, str(response.content))
+
+    def test_draft_menu_in_preview_mode(self):
+        """
+        Ensure that a draft page renders a draft menu when it exists.
+        """
+        menu_cont_published = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
+        pagecontent_draft = factories.PageContentWithVersionFactory(
```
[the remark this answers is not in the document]
Aiky30: With this current setup if you did `pagecontent_published.versions.count()` you would get 1, `pagecontent_published.content.page` would also be a different page to pagecontent_draft.content.psge

Aiky30 (changes requested):
The tests are not actually testing a page that has both a draft and published version.

We may have also broken the cache, I'll need to see the cache tests fail to know the correct way forward.

tests/test_plugin.py
```diff
@@ -407,43 +427,58 @@ def test_menu_cache_invalidate_after_menucontent_archive(self):
         # NOTE: This test is based on a similar one from django-cms:
         # https://github.com/divio/django-cms/blob/2daeb7d63cb5fee49575a834d0f23669ce46144e/cms/tests/test_plugins.py#L160
 
-        # Set up a versioned page with one placeholder
-        page_content = factories.PageContentWithVersionFactory(
-            language=self.language, version__created_by=self.get_superuser()
+        menu = factories.MenuFactory()
```
Aiky30, line 430: This test is now broken, they are very different to the other tests because they are testing logic that happens when you actually publish. The factories fake the transition of Draft -> Published -> Draft by just getting to the state directly. When you actually .publish() you are running the logic that happens on publish. It's important that this test works as it did before by running the publish method.

Aiky30: There should be one menucontent and one pagecontent for this test. Menu content will start published because we are testing the page publishing and the menu archiving!

```
menucontent = factories.MenuContentWithVersionFactory(language=self.language, version__state=PUBLISHED,)
pagecontent = factories.PageContentWithVersionFactory(language=self.language, version__created_by=self.get_superuser())
```

Aiky30 (changes requested):

tests/test_cms_menus.py
```diff
@@ -250,6 +252,145 @@ def test_get_roots_with_versioning_disabled(self):
...
+
+        self.assertEqual(pagecontent_draft.page, pagecontent_published.page)
+        self.assertEqual(menu_cont_draft.menu, menu_cont_published.menu)
+
+        # Node added in draft menu version is not rendered in published view, only published menu nodes are rendered
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.get(pagecontent_published.page.get_absolute_url())
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(pagecontent_draft.page, pagecontent_published.page)
```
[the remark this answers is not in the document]
NarenderRajuB: Updated

Aiky30 (approved):
All looks good now. :-)

PR #91, opened 2020-10-15
OPENCMS-950, OPENCMS-1083 Fix added for softroot navigation to remove it from navigation menu

NarenderRajuB:
Fix added to hide the softroot node in menu when hidden softroot menu item is rendered

### DIFF
--- a/tests/test_cms_menus.py
+++ b/tests/test_cms_menus.py
@@ -836,7 +836,7 @@ class SoftrootTests(CMSTestCase):
 
         self.assertTreeQuality(soft_root, mock_tree, 'level', 'title')
 
-    def test_menu_with_softroot_page_rendering(self):
+    def test_menu_with_hiddden_softroot_page_rendering(self):
         """
         Tree in fixture :
                root
@@ -848,10 +848,7 @@ class SoftrootTests(CMSTestCase):
                    bbb
         tag: show_menu 0 100 100 100
         expected result when rendering softroot node aaa:
-                     1:aaa1
-                        2:ccc
-                           3:ddd
-                     3:aaa2
+         : renders just the softroot page, with no menu nodes
         """
         menu_content_ver = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
         root = factories.ChildMenuItemFactory(parent=menu_content_ver.root, content=self.root_pagecontent.page)
--- a/tests/test_cms_menus.py
+++ b/tests/test_cms_menus.py
@@ -848,7 +848,9 @@ class SoftrootTests(CMSTestCase):
                    bbb
         tag: show_menu 0 100 100 100
         expected result:
-            renders just the softroot page, with no menu nodes
+            renders just the softroot page, with no menu nodes. It renders no nodes because the soft_root
+            and hide_node are True and the requirement is that when rendering a node that is soft_root and
+            hide_node the navigation should be empty
         """
         menu_content_ver = factories.MenuContentWithVersionFactory(version__state=PUBLISHED, language=self.language)
         root = factories.ChildMenuItemFactory(parent=menu_content_ver.root, content=self.root_pagecontent.page)
@@ -884,7 +886,8 @@ class SoftrootTests(CMSTestCase):
                        aaa2
                    bbb
         tag: show_menu 0 100 100 100
-        expected result:
+        expected result  when hiding a parent node the child node should not render a parent node that is
+        soft_root and hidden:
                     1:aaa1
                         2:ccc
                            3:ddd
--- a/tests/test_cms_menus.py
+++ b/tests/test_cms_menus.py
@@ -701,7 +701,6 @@ class SoftrootTests(CMSTestCase):
             parent=root,
             content=self.aaa_pagecontent.page,
             soft_root=True,
-            hide_node=True,
         )
         aaa1 = factories.ChildMenuItemFactory(parent=aaa, content=self.aaa1_pagecontent.page)
         ccc = factories.ChildMenuItemFactory(parent=aaa1, content=self.ccc_pagecontent.page)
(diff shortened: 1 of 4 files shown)


tests/test_cms_menus.py
```diff
@@ -839,6 +836,91 @@ def test_menu_with_softroot_page_rendering(self):
...
+               root
+                   aaa (soft_root)
+                       aaa1
+                           ccc
+                               ddd
+                       aaa2
+                   bbb
+        tag: show_menu 0 100 100 100
+        expected result when rendering softroot node aaa:
+         : renders just the softroot page, with no menu nodes
```
Aiky30, line 851: I don't understand this, if the softroot is aaa and we render aaa then it should show the children

tests/test_cms_menus.py
```diff
@@ -839,6 +836,91 @@ def test_menu_with_softroot_page_rendering(self):
...
+               root
+                   aaa (soft_root)
+                       aaa1
+                           ccc
+                               ddd
+                       aaa2
+                   bbb
+        tag: show_menu 0 100 100 100
+        expected result when rendering child node ccc of hidden softroot node aaa:
+         :          1:aaa1
```
Aiky30, line 888: Random : character here

tests/test_cms_menus.py
Aiky30: Also random : here

Aiky30 (changes requested):

tests/test_cms_menus.py
NarenderRajuB: > I don't understand this, if the softroot is aaa and we render aaa then it should show the children

Here when we render a softroot which is hidden , it shouldnot render its children, it will render softroot and children if the node is visible

tests/test_cms_menus.py
```diff
@@ -839,6 +836,91 @@ def test_menu_with_softroot_page_rendering(self):
...
+               root
+                   aaa (soft_root)
+                       aaa1
+                           ccc
+                               ddd
+                       aaa2
+                   bbb
+        tag: show_menu 0 100 100 100
+        expected result:
+            renders just the softroot page, with no menu nodes
```
Aiky30, line 851: We're still missing the why, if this broke I would have no idea what I'd actually broke, jus that this test is failing for some reason. I suspect it renders no nodes because the soft_root and hide_node are True and the requirement is that when rendering a node that is soft_root and hide_node the navigation should be empty?

If the above is correct we also need to edit the above to: aa (soft_root, hide_node)

tests/test_cms_menus.py
```diff
@@ -839,6 +836,91 @@ def test_menu_with_softroot_page_rendering(self):
...
+        Tree in fixture :
+               root
+                   aaa (soft_root)
+                       aaa1
+                           ccc
+                               ddd
+                       aaa2
+                   bbb
+        tag: show_menu 0 100 100 100
+        expected result when rendering child node ccc of hidden softroot node aaa:
```
Aiky30, line 887: Why is this the expected result?

I'm guessing that it's because when hiding a parent node the child node should not render a parent node that is soft_root and hidden.

tests/test_cms_menus.py
```diff
@@ -839,6 +836,91 @@ def test_menu_with_softroot_page_rendering(self):
...
+        soft_root = context['children']
+        mock_tree = []
+
+        self.assertTreeQuality(soft_root, mock_tree)
+
+    def test_menu_rendering_child_of_hidden_softroot_page(self):
+        """
+        Tree in fixture :
+               root
+                   aaa (soft_root)
```
Aiky30, line 880:
```suggestion
                   aaa (soft_root, hide_node)
```

tests/test_cms_menus.py
```diff
@@ -839,6 +836,91 @@ def test_menu_with_softroot_page_rendering(self):
 
         self.assertTreeQuality(soft_root, mock_tree, 'level', 'title')
 
+    def test_menu_with_hidden_softroot_page_rendering(self):
+        """
+        Tree in fixture :
+               root
+                   aaa (soft_root)
```
Aiky30, line 843:
```suggestion
                   aaa (soft_root, hide_node)
```

Aiky30 (changes requested):
Simply revising the comments in the tests will ensure that anyone who change the logic int he future and break a test understands what they have broken and its purpose (why). It's important to do this for complex features that will break easily with the smallest of logic changes.

tests/test_cms_menus.py
NarenderRajuB: yes you are right when rendering node that is soft_root and hide_node Navigation is empty.

tests/test_cms_menus.py
```diff
@@ -839,6 +836,91 @@ def test_menu_with_softroot_page_rendering(self):
...
+               root
+                   aaa (soft_root,hide_node)
+                       aaa1
+                           ccc
+                               ddd
+                       aaa2
+                   bbb
+        tag: show_menu 0 100 100 100
+        expected result:
+            renders just the softroot page, with no menu nodes
```
Aiky30, line 851:
```suggestion
            renders just the softroot page, with no menu nodes. It renders no nodes because the soft_root and hide_node are True and the requirement is that when rendering a node that is soft_root and hide_node the navigation should be empty
```

tests/test_cms_menus.py
```diff
@@ -839,6 +836,91 @@ def test_menu_with_softroot_page_rendering(self):
...
+        Tree in fixture :
+               root
+                   aaa (soft_root,hide_node)
+                       aaa1
+                           ccc
+                               ddd
+                       aaa2
+                   bbb
+        tag: show_menu 0 100 100 100
+        expected result:
```
Aiky30, line 887:
```suggestion
        expected result, when hiding a parent node the child node should not render a parent node that is soft_root and hidden.:
```

Aiky30 (changes requested):
The following comments help us to understand why as the test can be changed over time, we need to know the reason why. It is only valuable when you break the test in the future, you shouldn't have to guess what it is doing. The test enforces the requirement.

Aiky30 (approved):

PR #102, opened 2021-12-24
feat: changes to enable expand/collapse of tree in changelist view

marktsavage:
initial changes


djangocms_navigation/templatetags/navigation_admin_tree.py
```diff
@@ -0,0 +1,74 @@
...
+
+admin_tree.get_collapse = get_collapse
+
+
+t = get_template('djangocms_navigation/admin/tree_change_list_results.html')
+admin_tree.register.inclusion_tag(t, takes_context=True)(admin_tree.result_tree)
+
+
+@admin_tree.register.simple_tag
+def treebeard_js():
```
Aiky30, line 53: I think that we can do this in the admin or if that doesn't work the changeform. It's been done like that in the djangocms-pageadmin project.

djangocms_navigation/templatetags/navigation_admin_tree.py
```diff
@@ -0,0 +1,74 @@
...
+
+register = template.Library()
+
+'''
+This module is simply for overwriting some of treebeard's admin_tree templatetag functions
+with djangocms_navigation specific values.
+'''
+
+
+def get_spacer(first, result):
```
Aiky30, line 18: So is this how treebeard does this currently i.e. you're replacing kind of like for like template tags?

Aiky30: I'm not a fan of having markup in template tags.

djangocms_navigation/views.py
```diff
@@ -98,3 +99,12 @@ def get_data(self):
                     queryset = queryset.filter(**options)
 
         return queryset
+
+
+class MessageStorageView(View):
```
Aiky30, line 104: Is this replaced by the fact that we are happy to let the page reload or is it still required?

djangocms_navigation/static/djangocms_navigation/js/navigation-tree-admin.js
```diff
@@ -0,0 +1,367 @@
+(function ($) {
+        ACTIVE_NODE_BG_COLOR = '#B7D7E8';
```
Aiky30, line 2: Appears to be double indented. Some FE files are 2 spaces as per most guidelines, I'm not particularly bothers as long as the files in a specific project are consistent.

djangocms_navigation/static/djangocms_navigation/js/navigation-tree-admin.js
```diff
@@ -0,0 +1,367 @@
...
+                        backgroundColor: RECENTLY_MOVED_FADEOUT
+                    }, RECENTLY_FADE_DURATION, function () {
+                        this.removeAttribute('style');
+                    });
+                });
+            }
+        });
+    })(django.jQuery);
+    
+    // http://stackoverflow.com/questions/190560/jquery-animate-backgroundcolor/2302005#2302005
```
Aiky30, line 318: We should add a why this block is here, if one day this is a library or built in we would want to know if it can be removed / replaced.

djangocms_navigation/templates/admin/djangocms_navigation/menuitem/change_list.html
```diff
@@ -1,6 +1,11 @@
 {# Used for MP and NS trees #}
 {% extends "admin/tree_change_list.html" %}
```
Aiky30, line 2: This is where we can add the scripts that are currently a template tag "treebeard_js"

Aiky30 (changes requested):
We can talk through this on Monday. It's much bigger than we all would have thought it was Im sure. A task shouldn't require this many changes, the BA's would have broken it down further. Good work and I'm sure the changes we make are going to be cosmetic.

djangocms_navigation/templatetags/navigation_admin_tree.py
marktsavage: Yes, totally agree. Treebeard encapsulates some markup in their template tags, so I'm overwriting specific snippets.

djangocms_navigation/views.py
marktsavage: Because of the changed default expand/collapse, I've had to avoid a page reload. So this endpoint is needed to query the django messages queue that Treebeard updates when it moves nodes.

djangocms_navigation/static/djangocms_navigation/js/navigation-tree-admin.js
marktsavage: Well spotted! Indentation in this file is four spaces - I've updated and made consistent throughout.

Bernardvdv (approved):

tests/test_admin.py
```diff
@@ -842,6 +843,73 @@ def test_menuitem_changelist_redirects_if_not_latest_version(self, mocked_messag
...
+        (like moving a tree node) without reloading page.
+        """
+        menu_content = factories.MenuContentWithVersionFactory()
+        messages_endpoint = reverse(
+            "admin:djangocms_navigation_menuitem_message_storage", args=(menu_content.id,)
+        )
+
+        response = self.client.get(messages_endpoint)
+
+        self.assertEqual(response.status_code, 200)
```
Aiky30, line 859: We should also check that the endpoint returns something, we've had issues in the past where the endpoint returns a 200 but the result is wrong.

tests/requirements/requirements_base.txt
```diff
@@ -1,4 +1,6 @@
 coverage
```
Aiky30, line 1: The entries in this file should be kept in alphabetic order.

Aiky30 (changes requested):
Looks really good. 2 minor changes. 

If getting the response from the message view is too much hassle then it's fine to push back to get this in, if it's a simple addition then please add it.

adam-murray (approved):

tests/test_admin.py
marktsavage: No problem, I can easily put in a dummy entry to add a message to django messages queue. Will mock up an update to a navigation treebeard node to test it.

PR #104, opened 2022-02-16
Feature: Move delete from dropdown actions to changelist action button

adam-murray:
Partially based on work done by @narender81 in https://github.com/FidelityInternational/djangocms-navigation/pull/87.  Closes https://github.com/FidelityInternational/djangocms-navigation/pull/87

Moved delete option to admin action on each MenuItem. 

This is complicated by the fact that there isn't an easy way to get the root of a tree, given a nested node. This a problem because we need to find the MenuContent, as it is versioned, and therefore where we should be checking delete permissions for a given MenuItem. We also need it for redirecting after interactions with the delete view. This is done by overriding the changelist template, which allows us to pass MenuContent id to the admin via the confirmation, therefore allowing us to use the information after confirmation, as described above.


djangocms_navigation/admin.py
```diff
@@ -24,6 +28,11 @@
 from .views import ContentObjectSelect2View, MenuContentPreviewView
 
 
+IS_POPUP_VAR = '_popup'
```
Bernardvdv, line 31: This could be imported so no need to redefine the var

from django.contrib.admin.options import IS_POPUP_VAR

https://github.com/django/django/blob/main/django/contrib/admin/options.py#L70

marktsavage (approved):

Bernardvdv (approved):

djangocms_navigation/admin.py
```diff
@@ -362,7 +426,7 @@ def get_queryset(self, request):
         return self.model().get_tree()
 
     def change_view(
-        self, request, object_id, menu_content_id=None, form_url="", extra_context=None
+            self, request, object_id, menu_content_id=None, form_url="", extra_context=None
```
Aiky30, line 429:
```suggestion
        self, request, object_id, menu_content_id=None, form_url="", extra_context=None
```
Some additional formatting added unrelated to the changes being made here.

djangocms_navigation/admin.py
```diff
@@ -23,6 +28,10 @@
 from .views import ContentObjectSelect2View, MenuContentPreviewView
 
 
+ROOT_MESSAGE = _(
```
Aiky30, line 31: This is never done anywhere else in the packages, the messages are normally added inline unless they are reused or optional as per LOCK_MESSAGE.

This constant can be deleted and the error made inline.

djangocms_navigation/admin.py
```diff
@@ -467,6 +535,79 @@ def changelist_view(self, request, menu_content_id=None, extra_context=None):
...
+                kwargs={"menu_content_id": menu_content_id},
+            )
+            extra_context["list_url"] = list_url
+            if self._versioning_enabled:
+                menu_content = get_object_or_404(
+                    self.menu_content_model._base_manager, id=menu_content_id
+                )
+                delete_perm = self.has_delete_permission(request, menu_content)
+                if not delete_perm:
+                    messages.error(request, LOCK_MESSAGE)
```
Aiky30, line 556: Seems to be a gap in the test coverage as this lock message is not checked in deletion, see comments in the tests below.

tests/test_admin.py
```diff
@@ -842,6 +842,88 @@ def test_menuitem_changelist_redirects_if_not_latest_version(self, mocked_messag
...
+        new_child = factories.ChildMenuItemFactory(parent=menu_content.root)
+
+        url = reverse(
+            "admin:djangocms_navigation_menuitem_delete", args=(menu_content.id, new_child.id,)
+        )
+        with self.login_user_context(user=self.get_staff_user_with_no_permissions()):
+            response = self.client.get(url, follow=True)
+
+        self.assertEqual(response.status_code, 403)
+        self.assertEqual(MenuItem._base_manager.count(), 3)
```
Aiky30, line 908: Check the message for LOCK_MESSAGE here

Aiky30 (changes requested):
Looks good, 2 minor changes required.

Bernardvdv (approved):

tests/test_admin.py
```diff
@@ -842,6 +843,188 @@ def test_menuitem_changelist_redirects_if_not_latest_version(self, mocked_messag
...
+        'content': 60,
+        'permissions': 60,
+    },
+)
+class MenuItemAdminDeleteViewTestCase(CMSTestCase):
+    def setUp(self):
+        self.user = self.get_superuser()
+        self.client.force_login(self.user)
+
+    def test_menuitem_delete_view_success_single_item(self):
```
Aiky30, line 859:
```suggestion
    def test_menuitem_delete_view_with_permission(self):
```

tests/test_admin.py
```diff
@@ -842,6 +843,188 @@ def test_menuitem_changelist_redirects_if_not_latest_version(self, mocked_messag
...
+        """
+        With appropriate permissions, the delete view handles both the deletion of single, and multiple nodes (when
```
Aiky30, line 861:
```suggestion
        With appropriate permissions, the delete view can delete
```

tests/test_admin.py
```diff
@@ -842,6 +843,188 @@ def test_menuitem_changelist_redirects_if_not_latest_version(self, mocked_messag
...
+        we have a node with child(ren)).
```
Aiky30, line 862:
```suggestion
```

Aiky30 (approved):

PR #115, opened 2022-03-11
Added delete confirmation template to resolve breadcrumbs

Bernardvdv:


adam-murray (changes requested):
Could do with a test to ensure valid urls are being rendered.

marktsavage (approved):

Aiky30 (changes requested):
Looks good. It needs to have a test to ensure that the breadcrumb is generated as expected i.e. the correct url is output in the breadcrumbs. Could be a part of the following test or a new test similar for this issue: https://github.com/FidelityInternational/djangocms-navigation/blob/79f6dc89721b7d03f3ed08a64e2e27c647d46da9/tests/test_admin.py#L859

tests/requirements/requirements_base.txt
```diff
@@ -1,8 +1,10 @@
+beautifulsoup4
 coverage
 django-app-helper
 flake8
 factory_boy
 isort
+lxml
```
Aiky30, line 7:
```suggestion
```

lxml won't be required with beautiful soup, it's one or the other unless you are explicitly using it.

tests/test_admin.py
```diff
@@ -1030,6 +1031,33 @@ def test_menuitem_delete_view_root(self):
...
+            )
+
+        # Get all the hrefs in the markup for the breadcrumbs div
+        soup = BeautifulSoup(confirmation_response.content, features="lxml")
+        breadcrumb_html = soup.find("div", class_="breadcrumbs")
+        for a in breadcrumb_html.find_all('a', href=True):
+            with self.login_user_context(self.user):
+                breadcrumb_url = a['href']
+                response = self.client.get(breadcrumb_url)
+                self.assertEqual(response.status_code, 200)
```
Aiky30, line 1059: Normally we would check that the url is what we expect, I guess not throwing a 404 is ok, although it could give a 200 and be the wrong url. Fine to leave as is.

Aiky30 (changes requested):
Just the unused lxml dependancy to remove.

tests/test_admin.py
Bernardvdv: @Aiky30 Updated the test to validate the content of each url. Also looks like we need lxml for the CI to pass.

tests/test_admin.py
```diff
@@ -1030,6 +1031,67 @@ def test_menuitem_delete_view_root(self):
...
+            self.assertContains(
+                breadcrumb_first,
+                "<h1>Site administration</h1>"
+            )
+            self.assertContains(
+                breadcrumb_first,
+                '<a href="/en/admin/djangocms_navigation/menucontent/" class="changelink">Change</a>'
+            )
+            breadcrumb_second = self.client.get(breadcrumb_url[1])
+            self.assertContains(breadcrumb_second,
```
adam-murray, line 1071: The formatting looks a little strange here.

tests/test_admin.py
```diff
@@ -1030,6 +1031,67 @@ def test_menuitem_delete_view_root(self):
...
+            self.assertContains(breadcrumb_third,
+                                '<a href="/en/admin/djangocms_navigation/menuitem/1/1/change/'
+                                '">%s</a></th>' % menu_content.root)
+            breadcrumb_fourth = self.client.get(breadcrumb_url[3])
+            self.assertContains(
+                breadcrumb_fourth,
+                "<h1>Edit Menu: %s</h1>" % menu_content.root
+            )
+            self.assertContains(confirmation_response, '<li>Menu item: %s</li>' % child.title)
+            self.assertContains(confirmation_response, '<p>Are you sure you want to delete the menu item '
```
adam-murray, line 1091: Here again, would we not use new lines after opening the brackets for consistency?

adam-murray (approved):

tests/test_admin.py
Bernardvdv: Fixed

tests/test_admin.py
Bernardvdv: Fixed

Aiky30 (approved):
All good.

I'd say maybe a bit overkill on the responses from each url, may be a pain to maintain as django changes things. No point reversing anything though as it's potentially adding safety to all.

PR #122, opened 2022-03-25
Effectively removing pagination for changelist view.

marktsavage:
Disabled the pagination in MenuItem changelist view by setting the list_per_page to max value. This was desired to allow the full result set to display on one page.


Aiky30 (changes requested):
Needs a changelog entry and a brief description in the PR description.

adam-murray (approved):

PR #135, opened 2022-08-05
feat: Add main navigation functionality to allow the use of more than one navigation for a site/language

michaeljcollinsuk:
Added main_navigation functionality, this allows for the use of multiple navigation objects with the same language/site combination, while still being able to uniquely identify the navigation considered to be the main one for that combination.

The reason behind this is that if one extends the app to allow multiple navigations, such as side navigations, that aren't necessarily the same as the top/main navigation, we get inconsistent behaviour with different navigations returned.


djangocms_navigation/cms_menus.py
```diff
@@ -48,9 +49,26 @@ def get_roots(self, request):
...
         return queryset
 
+    def get_main_navigation(self, menucontents):
+        """
+        Takes a queryset of MenuContent objects, filters to include only those where the related Menu object has been
+        flagged as main navigation, and returns the queryset.
+        If this queryset is empty, because no Menu has been marked as the main navigation, the original queryset is
+        returned unchanged.
+        """
+        # main_navigation = menucontents.filter(menu__main_navigation=True).order_by("-versions__modified").first()
```
adam-murray, line 66: Please remove commented out code.

djangocms_navigation/cms_menus.py
```diff
@@ -48,9 +49,26 @@ def get_roots(self, request):
...
+
+            if getattr(settings, "MAIN_NAVIGATION_ENABLED", False):
+                menucontents = self.get_main_navigation(menucontents=menucontents)
+
             queryset = queryset.filter(menucontent__in=menucontents)
         return queryset
 
+    def get_main_navigation(self, menucontents):
+        """
+        Takes a queryset of MenuContent objects, filters to include only those where the related Menu object has been
```
adam-murray, line 61: Would you be able to follow this format for params/returns in this docstring?
https://github.com/FidelityInternational/djangocms-navigation/blob/f10af485558b8dbc6d199cefa5846bd7a5aece57/djangocms_navigation/helpers.py#L14

Although this hasn't been incorporated throughout the package already, we should try to ensure any changes we make are pushing the highest quality possible. This isn't something we have discussed though, so I wouldn't have expected it to be picked up implicitly!

tests/test_cms_menus.py
```diff
@@ -1495,3 +1543,188 @@ def test_menu_with_multiple_languages(self):
...
+        self.first_menucontent_child = factories.ChildMenuItemFactory(parent=self.first_menucontent.root)
+        self.second_menucontent = factories.MenuContentWithVersionFactory(
+            language="en",
+            version__state=PUBLISHED,
+            menu__main_navigation=False,
+            menu__site_id=1,
+        )
+        self.second_menucontent_child = factories.ChildMenuItemFactory(parent=self.second_menucontent.root)
+
+    def _get_nav_from_response(self, response):
```
adam-murray, line 1633: These helpers may be useful for other tests in the future, could you add this to a utils file or something in the test_utils directory?

tests/test_cms_menus.py
```diff
@@ -1495,3 +1543,188 @@ def test_menu_with_multiple_languages(self):
...
+
+        self.assertEqual(result, all_menucontents)
+        self.assertEqual(all_menucontents.count(), result.count())
+
+    def test_multiple_main_navigation(self):
+        """
+        When multiple MenuContent objects are related to a Menu marked as the main navigation are found, all the
+        MenuContent objects should be returned
+        """
+        older = factories.MenuContentWithVersionFactory(language="en", menu__main_navigation=True)
```
adam-murray, line 1590: Perhaps a bit pedantic, but could we change these to older_menucontent etc. Although it's a simple test, in general it's more readable to be a little more verbose.

adam-murray (approved):

PR #136, opened 2022-08-10
Feat: Added Main Navigation Option

adam-murray:
Added main_navigation functionality, this allows for the use of multiple navigation objects with the same language/site combination, while still being able to uniquely identify the navigation considered to be the main one for that combination. 

The reason behind this is that if one extends the app to allow multiple navigations, such as side navigations, that aren't necessarily the same as the top/main navigation, we get inconsistent behaviour with different navigations returned.


djangocms_navigation/admin.py
```diff
@@ -474,6 +519,51 @@ def changelist_view(self, request, menu_content_id=None, extra_context=None):
...
+                object_id=menu_content_id,
+                set_main_url=reverse(
+                    "admin:{app}_{model}_main_navigation".format(
+                        app=self.model._meta.app_label,
+                        model=self.menu_content_model._meta.model_name,
+                    ),
+                    args=(menu_content_id,),
+                ),
+                back_url=changelist_url,
+                extra_context=menu_queryset.values_list("identifier", flat=True),
```
michaeljcollinsuk, line 548: Is `menu_queryset.values_list("identifier", flat=True)` simply being used to check if there is an existing main nav object, or are you using the values somewhere? Because if you dont need the values, using `menu_queryset.exists()` would do the same thing more efficiently and is more explicit IMO.

michaeljcollinsuk (changes requested):

djangocms_navigation/admin.py
```diff
@@ -474,6 +519,51 @@ def changelist_view(self, request, menu_content_id=None, extra_context=None):
...
+                menucontent=menu.identifier,
+            )
+            return render(
+                request, "admin/djangocms_navigation/main_navigation_confirmation.html", context
+            )
+        if request.method == "POST":
+            for obj in menu_queryset:
```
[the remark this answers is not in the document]
adam-murray: Great suggestion, added.

djangocms_navigation/admin.py
```diff
@@ -474,6 +519,51 @@ def changelist_view(self, request, menu_content_id=None, extra_context=None):
...
+        menu_content = get_object_or_404(
+            self.menu_content_model._base_manager, id=menu_content_id
+        )
+        menu = menu_content.menu
+        menu_queryset = self.menu_model._base_manager.filter(
+            site=menu.site,
+            main_navigation=True,
+            menucontent__language=menu_content.language,
+        )
+        if request.method != "POST":
```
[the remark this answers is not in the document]
adam-murray: Thanks, have refactored this more along the lines of what was described.

djangocms_navigation/admin.py
adam-murray: This is used in the confirmation template, I have refactored it to make it clearer what is going on here.

djangocms_navigation/admin.py
```diff
@@ -474,6 +519,51 @@ def changelist_view(self, request, menu_content_id=None, extra_context=None):
 
         return super().changelist_view(request, extra_context)
 
+    def set_main_navigation_view(self, request, menu_content_id):
+        """Sets the selected navigation as the main navigation, if possible,
+        and redirects to the changelist.
+        """
+        changelist_url = get_admin_url(self.menu_content_model, "changelist")
```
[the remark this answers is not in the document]
adam-murray: Added, with additional test coverage.

djangocms_navigation/admin.py
```diff
@@ -474,6 +519,58 @@ def changelist_view(self, request, menu_content_id=None, extra_context=None):
...
+            # If this is a POST method, then they have confirmed, change the main_navigation
+            menu_queryset.update(main_navigation=False)
+            menu.main_navigation = True
+            menu.save()
+            self.message_user(
+                request, _(f"You have set the navigation {menu.identifier} as the main navigation.")
+            )
+            return redirect(changelist_url)
+
+        # If this isn't a POST method, it is the first interaction, therefore redirect to confirmation
```
michaeljcollinsuk, line 553:
```suggestion
        # If this isn't a POST method, it is the first interaction, therefore render confirmation
```
Very minor thing but you are not redirecting here so comment could cause confusion

djangocms_navigation/templates/admin/djangocms_navigation/main_navigation_confirmation.html
```diff
@@ -0,0 +1,42 @@
...
+<h3>{{ object_name }}</h3>
+    <h4>{% blocktrans %}Are you sure you would like to set the menu: <u>{{ menucontent }}</u> as the main navigation?{% endblocktrans %}</h4>
+<div>
+{% if extra_context.existing_menus %}
+    <h4>By doing this, the existing navigation below will no longer be the main navigation.</h4>
+<br>
+<ul>
+    {% for identifier in extra_context.existing_menus %}
+        <li>
+        <p>{{identifier}}</p>
```
michaeljcollinsuk, line 25:
```suggestion
        <p>{{ identifier }}</p>
```
Sorry to be picky but for consistency I would add spacing

michaeljcollinsuk (changes requested):

PR #138, opened 2022-08-17
feat: display navigation tree on preview view

michaeljcollinsuk:
### Description
Implements an improved preview view for a `MenuContent` and its related `MenuItem` objects. The preview view is based on the existing `changelist_view` but is intended as a "read only" version without the ability to add or change menu items, or view versions. Drag and drop is disabled, but the menu items are still collapsible so that you can better navigate the menu tree. 

Also implements a `edit` button to allow you to switch between preview and edit screens. The preview view is accessible regardless of the current version state, as was the case for the preview preview view.

This replaces the custom preview view that was served for the preview url, along with related tests. It also removes the previous admin `preview_view` and related URL that was not accessible.

### Checklist

- [x]  I have opened this pull request against master
- [x]  I have added or modified the tests when changing logic
- [x]  I have followed [the conventional commits guidelines](https://www.conventionalcommits.org/) to add meaningful information into the changelog


djangocms_navigation/static/djangocms_navigation/js/navigation-tree-admin.js
```diff
@@ -132,199 +132,203 @@ Original code found in treebeard-admin.js
...
-        // Activate all rows for drag & drop
-        // then bind mouse down event
-        $('td.drag-handler span').addClass('active').bind('mousedown', function (evt) {
-            $ghost = $('<div id="ghost"></div>');
-            $drag_line = $('<div id="drag_line"><span></span></div>');
-            $ghost.appendTo($body);
-            $drag_line.appendTo($body);
+        // Drag/drop only can be disabled. This ensures collapsible links are still activated.
+        // This is disabled when viewing a preview of a menu.
+        if ($('#disable-drag-drop').val() !== "1") {
```
michaeljcollinsuk, line 137: Note: this appears as a much bigger change than it is, it is simply adding this conditional which required the rest of the code to be outdented - the actual code is unchanged.

Bernardvdv (approved):
Looks really good!

djangocms_navigation/admin.py
```diff
@@ -476,33 +469,33 @@ def add_view(self, request, menu_content_id=None, form_url="", extra_context=Non
...
-            )
-        if request.GET.get("content_type_id"):
-            return True
-        else:
-            return False
+        request.menu_content_id = menu_content_id
+        menu_content = get_object_or_404(
+            self.menu_content_model._base_manager, id=menu_content_id
+        )
+        extra_context["title"] = "Preview Menu: {}".format(menu_content.__str__())
```
adam-murray, line 483: How come we are using .format here? This admin file already uses f strings, so I'd recommend using those.

tests/test_admin.py
```diff
@@ -301,6 +301,54 @@ def test_get_queryset_filters_by_content_id(self):
...
+        """
+        Check the template is the standard change list template
+        """
+        request = self.get_request("/admin/djangocms_navigation/menuitem/1/")
+        result = self.model_admin.get_changelist_template(request=request)
+        self.assertEqual(result, "admin/djangocms_navigation/menuitem/change_list.html")
+
+    def test_get_changelist_template_preview(self):
+        """
+        Check the template is the preview template
```
adam-murray, line 346:
```suggestion
        Check the preview template is used when hitting the preview endpoint.
```

tests/test_admin.py
```diff
@@ -301,6 +301,54 @@ def test_get_queryset_filters_by_content_id(self):
...
+        mock_list_actions.assert_called_once_with(request)
+        self.assertEqual(len(list_display), 5)
+        self.assertEqual(
+            list_display,
+            ['__str__', 'get_object_url', 'soft_root', 'hide_node', "list_actions"]
+        )
+
+    def test_get_changelist_template(self):
+        """
+        Check the template is the standard change list template
```
adam-murray, line 338: Check the template is the standard changelist template when accessing the changelist.

tests/test_admin.py
```diff
@@ -809,6 +861,119 @@ def test_menuitem_add_view_redirects_on_save_add_another(self):
...
+
+        self.assertEqual(response.status_code, 200)
+        self.assertNotContains(response, "Versions")
+        self.assertNotContains(response, version_url)
+
+    def test_actions_not_present(self):
+        """
+        Check that the Actions column is not present
+        """
+        response = self.client.get(self.preview_url)
```
adam-murray, line 911: Could you add a new line after this, so that the format is consistent with the other tests?

tests/test_admin.py
```diff
@@ -809,6 +861,119 @@ def test_menuitem_add_view_redirects_on_save_add_another(self):
...
+        """
+        Check that the expand / collapse all controls are present.
+        """
+        response = self.client.get(self.preview_url)
+
+        soup = BeautifulSoup(str(response.content), features="lxml")
+        element = soup.find("th", class_="expand-all")
+        link = element.find("a")
+
+        self.assertIsNotNone(element)
```
adam-murray, line 959: Would an empty list not always be not None?

adam-murray (changes requested):
Really good work mate, I've added a couple of comments, but had to be really pedantic to find anything, great job!

djangocms_navigation/admin.py
michaeljcollinsuk: It was because it was lifted from the [changelist_view](https://github.com/django-cms/django-cms/commit/ff6cb9b5dced92eadef62694e989d601e9475b30) which uses `.format`, I hadn't noticed the file using f-strings but I personally prefer it for most circumstances so happy to change.

tests/test_admin.py
```diff
@@ -301,6 +301,54 @@ def test_get_queryset_filters_by_content_id(self):
...
+            list_display,
+            ['__str__', 'get_object_url', 'soft_root', 'hide_node', "list_actions"]
+        )
+
+    def test_get_changelist_template(self):
+        """
+        Check the template is the standard change list template when the request is for the changelist endpoint
+        """
+        request = self.get_request("/admin/djangocms_navigation/menuitem/1/")
+        result = self.model_admin.get_changelist_template(request=request)
```
adam-murray, line 341: Would we be able to have a new line between the Arrange, Act and Assert in this test?

tests/test_admin.py
```diff
@@ -301,6 +301,54 @@ def test_get_queryset_filters_by_content_id(self):
...
+        self.assertEqual(result, "admin/djangocms_navigation/menuitem/change_list.html")
+
+    def test_get_changelist_template_preview(self):
+        """
+        Check the preview template is used when the request is for the preview endpoint
+        """
```
adam-murray, line 347: Same for AAA here please!

tests/test_admin.py
```diff
@@ -809,6 +861,120 @@ def test_menuitem_add_view_redirects_on_save_add_another(self):
...
+        )
+
+    def test_invalid_menucontent_id_raises_404(self):
+        """
+        Check that 404 response is returned if the menu_content_id is invalid
+        """
+        url = reverse("admin:djangocms_navigation_menuitem_preview", args=(999,))
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 404)
```
adam-murray, line 881: Could we have a new line so this fits AAA?

tests/test_admin.py
```diff
@@ -809,6 +861,120 @@ def test_menuitem_add_view_redirects_on_save_add_another(self):
...
+    def test_menuitem_preview_context(self):
+        """
+        Check that the context contains the correct title and the menucontent object
+        """
+        model_admin = MenuItemAdmin(MenuItem, admin.AdminSite())
+        request = self.get_request("/admin/djangocms_navigation/menuitem/1/preview/")
+        request.user = self.get_superuser()
+        menu_content = factories.MenuContentWithVersionFactory()
+
+        response = model_admin.preview_view(request=request, menu_content_id=menu_content.pk)
```
adam-murray, line 973: Could we have a new line here as well, for AAA.

adam-murray (approved):

PR #140, opened 2022-08-22
Feature/add menu item distinct nodes

michaeljcollinsuk:
### Description
Fixes a bug that was causing duplicate objects to be displayed for selection when a user adds a menu item node.

In addition the Page objects returned for selection are filtered by the current site and language.

As part of this, the view has been refactored to simplify the logic used to get data and remove readability by removing nested/excessive conditionals. This includes removing the hardcoded filter on the title for `Page` objects in order to make use of the `navigation_models` attribute on the app config as it was intended. As such, the changes include extensive additional unit tests for the `ContentObjectSelect2View.get_data` method.

I believe that the `get_data` could be refactored further, as some of the implemented functionality is unused outside of tests. However I decided this was out of scope for the current ticket, but is something we could revisit in future.

### Checklist

- [x]  I have opened this pull request against master
- [x]  I have added or modified the tests when changing logic
- [x]  I have followed [the conventional commits guidelines](https://www.conventionalcommits.org/) to add meaningful information into the changelog

### DIFF
--- a/djangocms_navigation/views.py
+++ b/djangocms_navigation/views.py
@@ -64,26 +64,23 @@ class ContentObjectSelect2View(View):
         if pk:
             queryset = queryset.filter(pk=pk)
 
-        if query:
+        if not query:
+            return queryset
 
-            # TODO: filter by language and publish state
-            # For Page model filter query by pagecontent title
-            if model == Page:
-                queryset = queryset.filter(pagecontent_set__title__icontains=query)
-            else:
-                # Non page model should work using filter against field in queryset
-                options = {}
-                search_fields = supported_models(self.menu_content_model).get(model)
-                if search_fields:
-                    for field in search_fields:
-                        options[field] = query
-                    queryset = queryset.filter(**options)
-            # if we have a search query we call distinct as for a Page this results in an SQL join due to the filter
-            # being across the reverse FK relation to pagecontent_set. This may also be the case for other content types
-            # so the distinct() call is included whenever we have a search query.
-            queryset = queryset.distinct()
+        # TODO: filter by language and publish state
+        # For Page model filter query by pagecontent title
+        if model == Page:
+            return queryset.filter(pagecontent_set__title__icontains=query).distinct()
 
-        return queryset
+        # Non page model should work using filter against field in queryset
+        search_fields = supported_models(self.menu_content_model).get(model)
+        # if no fields to filter against were declared return the queryset unchanged
+        if not search_fields:
+            return queryset
+
+        options = {field: query for field in search_fields}
+        # the filter could be across tables so distinct should be used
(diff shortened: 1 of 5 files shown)


djangocms_navigation/views.py
```diff
@@ -64,22 +64,23 @@ def get_data(self):
         if pk:
```
michaeljcollinsuk, line 64: I didnt want to refactor too heavily here, but I did wonder if this could be simplified further by removing other logic such as this, as it does not appear to be used.

djangocms_navigation/views.py
```diff
@@ -64,22 +64,23 @@ def get_data(self):
...
-                if search_fields:
-                    for field in search_fields:
-                        options[field] = query
-                    queryset = queryset.filter(**options)
-
-        return queryset
+        if not query:
+            return queryset
+
+        # TODO: filter by language and publish state
```
michaeljcollinsuk, line 70: Should this still be implemented?

djangocms_navigation/views.py
```diff
@@ -55,31 +57,27 @@ def get_data(self):
         except (TypeError, ValueError):
             pk = None
 
-        if site:
-            if hasattr(model.objects, "on_site"):
```
michaeljcollinsuk, line 59: I have had to change this as previous implementation was causing an error. This is because even when the content type is a `Page`, the queryset returned on line 49 (`queryset = versionable_item.grouper_choices_queryset()`) is a standard `QuerySet` instance rather than a `PageQuerySet` instance.

[The `on_site` method exists on `PageQueryset`](https://github.com/django-cms/django-cms/blob/ff6cb9b5dced92eadef62694e989d601e9475b30/cms/models/query.py#L10), which would result in the conditional returning true. But then actual call was being made on a standard `QuerySet` instance where `on_site` does not exist, resulting in a `AttributeError`.

tests/test_views.py
```diff
@@ -214,3 +224,243 @@ def test_select2_poll_content_view_pk(self):
...
+
+                expected = getattr(self, language)
+                self.assertEqual(Page._base_manager.count(), 4)
+                self.assertEqual(results.count(), 1)
+                self.assertIn(expected.page, results)
+
+    @patch("djangocms_navigation.views.get_current_site")
+    def test_page_queryset_filtered_by_site(self, mock_get_current_site):
+        """
+        Check that when pages belong to different sites, only sites for
```
Bernardvdv, line 437: Not sure if this is docstring is complete?

michaeljcollinsuk: Oops, fixed

Bernardvdv (approved):

adam-murray (approved):
Excellent work mate!

PR #143, opened 2022-09-12
feat: filter by PageUrl when adding menu node to navigation

michaeljcollinsuk:
Updates filter query so that when searching for a node by multiple fields, the filter query uses the OR operator rather than AND. This allows searching by multiple fields.

Updates the CMS config so the slug and the path of the PageUrl is used to filter the Page queryset when adding a node to a navigation menu.


adam-murray (approved):
Great work mate!

Bernardvdv (approved):

PR #145, opened 2022-09-13
fix: Navigation nodes with page display edit links in edit mode.

adam-murray:
Navigation nodes with page attached now display edit urls when in edit mode


adam-murray:
Converting to draft until confirmation that this is not required is given.

PR #146, opened 2022-09-15
fix: return main navigation for the current site

michaeljcollinsuk:
Fixes a bug where a Menu marked as the main navigation for a different site could be returned when filtering for a main navigation menu. This would result no root nodes being returned, which would cause the application to time out while attempting to load the navigation menu.

### DIFF
--- a/djangocms_navigation/cms_menus.py
+++ b/djangocms_navigation/cms_menus.py
@@ -37,8 +37,9 @@ class CMSMenu(Menu):
 
     def get_roots(self, request):
         language = get_language_from_request(request)
+        site = get_current_site()
         queryset = self.menu_item_model.get_root_nodes().filter(
-            menucontent__menu__site=get_current_site()
+            menucontent__menu__site=site
         )
         versionable = get_versionable_for_content(self.menu_content_model)
         if versionable:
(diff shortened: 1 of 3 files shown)


djangocms_navigation/cms_menus.py
```diff
@@ -65,7 +66,7 @@ def get_main_navigation(self, menucontents):
 
         :param menucontents: A QuerySet of MenuContent instances
         """
```
Bernardvdv, line 68:
```suggestion
        :param site: Current site object
        """
```

Bernardvdv (commented):
Just missing a changelog and the minor param change. Looks good

Bernardvdv (approved):

tests/test_cms_menus.py
```diff
@@ -1595,12 +1598,76 @@ def test_multiple_main_navigation(self):
...
+
+        result = self.menu.get_main_navigation(menucontents=all_menucontents, site=site_1)
+
+        self.assertEqual(result.count(), 2)
+        self.assertIn(site_1_menu, result)
+        self.assertIn(site_2_menu, result)
+
+    def test_no_main_navigation_for_current_site(self):
+        """
+        Test that with multiple sites, and the current site has no main navigation, and another site
```
adam-murray, line 1652:
```suggestion
        With the current site having no main navigation, and another site having a main navigation, we return the navigation from the current site.
```

adam-murray (approved):

PR #150, opened 2022-09-20
feat: moving a menu item node requires user confirmation

michaeljcollinsuk:
Uses javascript confirm function to display a dialog box when a user moves a menu item node.

### DIFF
--- a/djangocms_navigation/static/djangocms_navigation/js/navigation-tree-admin.js
+++ b/djangocms_navigation/static/djangocms_navigation/js/navigation-tree-admin.js
@@ -230,7 +230,10 @@ Original code found in treebeard-admin.js
                         }
                     });
                 }).bind('mouseup',function () {
-                    if ($targetRow !== null) {
+                    // prompt user to confirm the move
+                    let name = node.node_name()
+                    let confirmMove = confirm(`Are you sure you want to move menu item ${name}?`)
+                    if ($targetRow !== null && confirmMove === true) {
                         target_node = new Node($targetRow[0]);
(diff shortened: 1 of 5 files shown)


djangocms_navigation/static/djangocms_navigation/js/navigation-tree-admin.js
```diff
@@ -230,7 +230,10 @@ Original code found in treebeard-admin.js
                         }
                     });
                 }).bind('mouseup',function () {
-                    if ($targetRow !== null) {
+                    // prompt user to confirm the move
+                    let name = node.node_name()
+                    let confirmMove = confirm(`Are you sure you want to move menu item ${name}?`)
```
Bernardvdv, line 235: Will this be translatable? If not, can we define the confirm message in the backend?

michaeljcollinsuk: I hadn't considered this, I think it will need to be done on the backend. Do you use [django translation](https://docs.djangoproject.com/en/4.1/topics/i18n/translation/) for this elsewhere?

Bernardvdv: We had a similar situation with filer previously where we had to configure a message in JS. Hope this helps :-) 

https://github.com/divio/djangocms-versioning-filer/pull/89/files#diff-04d9554189848fde88c8b80f5870729b4790cc953a92e8c8be8330aa0cb31e6fR103

adam-murray (approved):

Bernardvdv (approved):

PR #153, opened 2022-10-12
Feature/expanded nodes are stored in the page session so that the structure of the nav tree is retained after editing a menu

michaeljcollinsuk:
Expanded nodes are stored in the page session so that the structure of the nav tree is retained after editing a menu.

Stores a list of node ID's in a JS page session object so that after editing a menu, or deleting a node, the menu is displayed in the same structure as before beginning the change, with the correct nodes expanded. The session object expires when the browser page or tab is closed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,8 @@ Changelog
 
 Unreleased
 ==========
+* feat: store a list of expanded nodes in a page session object so after editing a menu the
+previous state is displayed
 
 1.7.0 (2022-09-21)
 ==================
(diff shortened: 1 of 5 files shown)


Bernardvdv (commented):
Changes looks good. My only comment is there is a mixture of ES6 and its predecessors. Not sure if it is worth converting the whole file to ES6?

michaeljcollinsuk:
> Changes looks good. My only comment is there is a mixture of ES6 and its predecessors. Not sure if it is worth converting the whole file to ES6?

Or perhaps better not to use ES6 to keep it consistent? Would just be a little concerned about unknowingly breaking something if I edit the whole file

Bernardvdv (approved):
Approving on the base that the file can be converted to ES6 at a later stage

adam-murray (approved):

PR #155, opened 2022-10-20
fix: display node tree on menu item delete confirmation page

michaeljcollinsuk:
Fixes a bug which meant not all child nodes were displayed on the delete confirmation page. Also ensures that the node tree displays items with the correct nesting to reflect their position in the tree.


Bernardvdv (approved):

adam-murray (approved):
Just needs brought up to date. Great work!

PR #159, opened 2023-02-01
feature removed unpublished pages from menu item admin select view fo…

vipulnarang95:
Removed unpublished pages from menu item select view admin

### DIFF
--- a/djangocms_navigation/views.py
+++ b/djangocms_navigation/views.py
@@ -37,7 +37,9 @@ class ContentObjectSelect2View(View):
         # Removing unpublished pages from queryset
         if model == Page:
             queryset_data = [
-                page for page in queryset_data if page.get_title_obj().versions.first().state != 'unpublished'
+                page for page in queryset_data
+                if getattr(page.get_title_obj().versions.first(), "state", None)
+                != 'unpublished'
             ]
 
         data = {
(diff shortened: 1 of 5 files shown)


djangocms_navigation/views.py
```diff
@@ -32,8 +32,18 @@ def get(self, request, *args, **kwargs):
         if not is_model_supported(self.menu_content_model, model):
             return HttpResponseBadRequest()
 
+        queryset_data = self.get_data()
+
+        # Removing unpublished pages from queryset
+        if model == Page:
+            queryset_data = [
+                page for page in queryset_data
+                if getattr(page.get_title_obj().versions.first(), "state", None)
```
joshyu, line 41: add `from djangocms_versioning.constants import ARCHIVED, UNPUBLISHED` at the top of the page.

update below codes to include checking  `archived`.
`
if not getattr(page.get_title_obj().versions.first(), "state", None)
in [ ARCHIVED, UNPUBLISHED ]
`

If you check the unpublished page '**Triage tool test**'  in `EU WPFH Multisite NP`, you can find the `get_title_obj` will return previously published pagecontent instance (see below screenshot), but now its state is `archived`. so I think that return of `archived` page is also meaningless. 

image

vipulnarang95: @joshyu Can you please check now?

joshyu (approved):

